PR #2, opened 2021-08-16
Refactor Register Relay

JONAF2103:
## Introduction
The Relay Server was polling the events from the chain to detect the registration process. That has some security issues and also is not the best solution, with this refactor we are querying the chain from a view directly into a mapping. This solution is better in terms of security and performance.

## Changes

This PR changes the Relay Hub contract to add the relay data mapping where the relay servers are registered.


wilsoniovlabs (approved):
LGTM

mortelli (approved):

antomor (approved):

wilsoniovlabs (approved):
LGTM

PR #5, opened 2021-09-09
Update dist

antomor:
- Update the dist folder


antomor:
@mortelli Could you please review it?

mortelli (commented):
i'm only finding one difference here and that's the `dist/idex.js` file. but since this file is minified and over 3 MB long (which is a lot for text) i'm not really sure of what's up.

could you please try `rm -rf build/ dist/ package-lock.json && npm i && npm run dist` and see if you get any diffs?

if not let's take a look together :mag: 

thanks

mortelli (approved):
it seems `index.js` includes some sort of time-bases data with each compilation (maybe a timestamp). this will cause it to be different every single time it's built.

therefore, i suggest we only upload changes to this file as long as other files are changed alongside of it.

approving

PR #8, opened 2021-09-16
Adding package lock file

wilsoniovlabs:


mortelli (commented):
i agree on the changes so far, but i think we still have some things to resolve before closing this.

as @antomor suggested on JIRA:
> We need to investigate further on the --package-lock-only parameter to avoid committing package-lock.json / yarn.lock file.
>
> If that’s not feasible, it could be necessary to create and commit one between the package-lock.json and yarn.lock files.

in addition to this, i'd like to pose the following question:

if i download this branch and execute `rm -rf node_modules/ dist/ build/ && npm i && npm run build && npm run dist` i get differences in the `package-lock.json` file. 

do you know why this is and do you have any proposals on how to handle it?

mortelli:
> i agree on the changes so far, but i think we still have some things to resolve before closing this.
> 
> as @antomor suggested on JIRA:
> 
> > We need to investigate further on the --package-lock-only parameter to avoid committing package-lock.json / yarn.lock file.
> > If that’s not feasible, it could be necessary to create and commit one between the package-lock.json and yarn.lock files.
> 
> in addition to this, i'd like to pose the following question:
> 
> if i download this branch and execute `rm -rf node_modules/ dist/ build/ && npm i && npm run build && npm run dist` i get differences in the `package-lock.json` file.
> 
> do you know why this is and do you have any proposals on how to handle it?

same goes for:
- https://github.com/infuy/rif-relay-server/pull/8
- https://github.com/infuy/rif-relay-common/pull/7
- https://github.com/infuy/rif-relay-client/pull/7

antomor:
Yep, as previously mentioned, I think we should think about our use cases to better understand what to do. Although we could commit the lock files to avoid problems at the beginning, in the long run, it could be counterproductive, especially if we think that the library will be installed by other users.
I have just updated the JIRA issue, to include some contrasting opinions on that.

package.json
```diff
@@ -18,36 +18,37 @@
     "allowTokens": "./scripts/allowTokens",
     "allowedTokens": "./scripts/allowedTokens",
     "prettier:fix": "npx prettier --write --ignore-unknown .",
-    "prettier": "npx prettier --check ."
+    "prettier": "npx prettier --check .",
+    "switchPostInstall": "scripts/switchPostInstall"
```
antomor, line 22: why do we need the `switchPostInstall` script?

wilsoniovlabs: I recovered this change for mistake.
Removed.

PR #9, opened 2021-10-28
PP-43: corrections to Refactor Register Relay

mortelli:
## What

- rename `IRelayHub.RelayData` to `IRelayHub.RelayManagerData`
- replace the `RelayManagerData.stakeAdded` & `RelayManagerData.penalized` fields with `RelayManagerData.currentlyStaked`
- remove some extra `require` statements in Relay Hub contract

## Why

- code inspection has shown some errors in the current implementation

## Refs

- [PP-43](https://rsklabs.atlassian.net/browse/PP-43)
- [Refactor Register Relay: changes to be implemented](https://hackmd.io/XS3gcRwjQxe9qetWQ4o8vw?view)


contracts/RelayHub.sol
```diff
@@ -35,7 +35,7 @@ contract RelayHub is IRelayHub {
     mapping(address => StakeInfo) public stakes;
 
     //maps relay manager address to the relay data
-    mapping(address => RelayData) public relayData;
+    mapping(address => RelayManagerData) public relayData;
```
mortelli, line 38: the name `RelayInfo` was preferable, but was already taken (`rif-relay-client/src/types/RelayInfo.ts`)

contracts/RelayHub.sol
```diff
@@ -285,16 +285,15 @@ contract RelayHub is IRelayHub {
...
+        relayData[relayManager].currentlyStaked = false;
 
         // RBTC is burned and transferred
         address(0).transfer(toBurn);
         beneficiary.transfer(reward);
         emit StakePenalized(relayManager, beneficiary, reward);
     }
 
-    function getRelayData(address relayManager) external view override returns (RelayData memory relayInfo) {
+    function getRelayInfo(address relayManager) external view override returns (RelayManagerData memory relayManagerData) {
```
mortelli, line 296: renamed to be structurally similar to `getStakeInfo`

contracts/RelayHub.sol
```diff
@@ -368,8 +368,6 @@ contract RelayHub is IRelayHub {
         require(info.owner == msg.sender, "not owner");
         require(info.withdrawBlock > 0, "Withdrawal is not scheduled");
         require(info.withdrawBlock <= block.number, "Withdrawal is not due");
-        require(relayData[relayManager].manager != address(0), "Relay is not registered");
-        relayData[relayManager].penalized = true;
```
mortelli, line 372: the logic for `getRelayInfoForManagers` checks for unlock state, not withdraw, so this was moved (see line `362` on the right)

contracts/interfaces/IRelayHub.sol
```diff
@@ -107,14 +107,13 @@ interface IRelayHub {
...
-    // get the relay data from the manager address
-    function getRelayData(address relayManager) external view returns(RelayData memory relayInfo);
+    // get the relay info from the manager address
+    function getRelayInfo(address relayManager) external view returns(RelayManagerData memory relayManagerData);
 
-    // Represents the relay data for the relay
-    struct RelayData {
+    // Represents the relay data for a particular relay manager
+    struct RelayManagerData {
         address manager;
```
mortelli, line 115: i would have liked to remove this field, but it's useful on the javascript side.

the reason why i'd like not to have it is that it's redundant: the key to access this map will have the same value as this field. 

however, once an entry is received from the javascript side we no longer have the mapping around, so it's good to keep this value at hand.

we'd also need a way to signal whether a specific key has a value in the solidity mapping, and an `exists` bool field isn't ideal

contracts/RelayHub.sol
```diff
@@ -285,16 +285,15 @@ contract RelayHub is IRelayHub {
         uint256 toBurn = SafeMath.div(amount, 2);
         uint256 reward = SafeMath.sub(amount, toBurn);
 
-        require(relayData[relayManager].manager != address(0), "Relay is not registered");
```
mortelli, line 288: this was removed because the old code did not enforce this check, so the refactor should not either

contracts/RelayHub.sol
```diff
@@ -368,8 +368,6 @@ contract RelayHub is IRelayHub {
         require(info.owner == msg.sender, "not owner");
         require(info.withdrawBlock > 0, "Withdrawal is not scheduled");
         require(info.withdrawBlock <= block.number, "Withdrawal is not due");
-        require(relayData[relayManager].manager != address(0), "Relay is not registered");
```
mortelli, line 371: this was removed because the old code did not enforce this check, so the refactor should not either

antomor (approved):
LVGTM!

PR #10, opened 2021-10-28
PP-98/Change the network name to network id

wilsoniovlabs:
## What
Revert change that replace the id of the network to the name. This change is on migrate and contract-addresses.json.

## Why
On relaying services SDK have error because they not get the contract list from rif-relay-contracts contract-addresses file.

## Refs
[Change the network name](https://rsklabs.atlassian.net/browse/PP-98)

Revert change on rif-relay-contracts to change the network name x the network id.
Modify the deploy migration and make this changes.

### DIFF
--- a/contract-addresses.json
+++ b/contract-addresses.json
@@ -1,5 +1,5 @@
 {
-    "regtest": {
+    "33": {
         "penalizer": "0xe8C7C6f18c3B9532343487faD807060750C1fE95",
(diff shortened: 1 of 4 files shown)


mortelli (approved):
both the allow tokens functionality as well as the tests seem to work OK, so i am approving.

tagging @antomor just in case he wants to take a look before merging (no obligation, up to you)

antomor (approved):
LVGTM!

PR #13, opened 2021-12-15
Add Testnet addresses

antomor:
## What

- Update the contract-addresses.json to include the Testnet deployment

## Why

- the latest version of the contracts were deployed on Testnet

## Refs

- [PP-95](https://rsklabs.atlassian.net/browse/PP-95)


mortelli (approved):
although i cannot see the code, i used the RSK testnet explorer to verify that each of these addresses correspond to a smart contract (plus the test token)

PR #14, opened 2022-01-06
revenue sharing POC

mortelli:
PR to merge work between myself and @antomor for the revenue sharing POC.


mortelli (commented):
there still needs to be some work done for the 3rd migration and the new scripts (mainly have them use the same variables set in the new migration, or from somewhere else if we move these)

i can lend a hand with this

tasks/create-safe.js
```diff
@@ -0,0 +1,282 @@
+const safeFactoryABI = [
```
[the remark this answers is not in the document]
antomor: done ;-)

antomor:
Maybe I missed something, but I think I included the changes you talk about

mortelli:
> Maybe I missed something, but I think I included the changes you talk about

thanks, will take a better look

tasks/utils.js
```diff
@@ -0,0 +1,48 @@
...
+        safeAddress: safeSdk.getAddress()
+    });
+    const txHash = await safeSdk2.getTransactionHash(safeTransaction);
+    const approveTxResponse = await safeSdk2.approveTransactionHash(txHash);
+    await approveTxResponse.transactionResponse?.wait();
+};
+
+const safeProxyFactoryAddress = '0x73ec81da0C72DD112e06c09A6ec03B5544d26F05';
+const multiSendAddress = '0x5159345aaB821172e795d56274D0f5FDFdC6aBD9';
+const safeMasterCopyAddress = '0x83C5541A6c8D2dBAD642f385d8d06Ca9B6C731ee';
```
mortelli, line 32: do we have any way of reading these from somewhere like a file?

mortelli, line 32: which contract is this one? i assume the first 2 are `GnosisSafeProxyFactory` and `MultiSend`. is the last one `GnosisSafe`?

tasks/send-multisig.js
```diff
@@ -0,0 +1,91 @@
...
+const { contractNetworks, signWithAddress } = require('./utils');
+const RevenueSharingAddresses = require('../revenue-sharing-addresses.json');
+
+const EMPTY_DATA = '0x';
+
+/**
+ * This script can be called with the following format:
+ * `npx truffle --network regtest exec tasks/send-multisig.js --safeAddress='0x3d9518b0852183c5c544a2161575208A6543D61C'`
+ * Please keep in mind the following assumptions:
+ * 1. the safe have two owners, that are the first two accounts (accounts[0], accounts[1])
```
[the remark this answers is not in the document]
antomor: Yes, you're right

tasks/utils.js
```diff
@@ -0,0 +1,48 @@
...
+    return testTokenInstance;
+};
+
+const getCollectorInstance = async (artifacts) => {
+    const Collector = artifacts.require('Collector');
+    const collectorInstance = await Collector.deployed();
+    return collectorInstance;
+};
+
+const signWithAddress = async (web3, safeSdk, safeTransaction, owner) => {
```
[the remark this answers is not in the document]
antomor: In the safe context, they are referenced as owners and yes, I believe the owner is more appropiate in that sense. Not all addresses can sign a safe transaction.

tasks/utils.js
antomor: Yes, it is. I'm going to change its name for clarity (within the SDK it's called like that)

antomor: Not sure, because right now we deploy the contracts from the safe repo directly

tasks/utils.js
```diff
@@ -0,0 +1,48 @@
...
+};
+
+module.exports = {
+    getTestTokenInstance,
+    getCollectorInstance,
+    signWithAddress,
+    contractNetworks,
+    safeMasterCopyAddress,
+    safeProxyFactoryAddress
+};
```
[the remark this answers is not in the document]
antomor: You're right, thank you for pointing it out :-)

tasks/utils.js
mortelli: alright, resolving then

mortelli: leave as-is then

PR #16, opened 2022-01-11
Fix the code that waits for the transaction receipt

antomor:
## What

- Change the code that waits for the transaction receipt. Although that code was retrieved from the official documentation, it doesn't work as expected resulting in timing-related errors.

## Why

- We want to wait for the transaction receipt to be available otherwise, unexpected errors are raised showing that the transactions were not signed by other owners.


mortelli (approved):
thanks for this. i've tested out several scenarios:
1. withdraw without balance
2. withdraw with balance sent through truffle
3. withdraw with balance sent through the dapp
4. successive withdrawals

no failures so far :+1:

PR #20, opened 2022-02-08
PP-123: Collector deployment task

antomor:
## What

- Remove the revenue sharing migration file
- Add a script to deploy a collector

## Why

- We didn't have the possibility to execute the migration script with some additional parameters (collector owners and partner properties)
- We want to separate the contract migrations from the collector deployment.


mortelli (approved):
superb work :+1: 

i tested this out manually and collectors are behaving like they should; i was able to call `withdraw` through truffle with correct results. i also tested out the task parameters (default or incorrect values) and all seems good.

a few comments:
1. if parameters are invalid for the deployment of a collector the revert reason is not shown, instead:

```
    const collectorInstance = await Collector.new(
                                                 ^
Error: The contract code couldn't be stored, please check your gas limit.
``` 
i'm not sure something can be done about this, but i'm mentioning it just in case.

2. do you think it's possible/beneficial to also have a token address parameter and not always use the `TestToken`?
3. this appears to break the `withdraw` script (as the `Collector` contract is no longer part of the migrations). i'm not too bothered by this, since a user would need more help deploying a collector than just calling a `withdraw`. however, i'd like to ask your opinion on what you think we should do about this file.

thanks!

antomor:
@mortelli Thanks for your review! 

> 2. do you think it's possible/beneficial to also have a token address parameter and not always use the TestToken?

Good catch, I've just fixed it to receive the token address as well.

> 3. this appears to break the withdraw script (as the Collector contract is no longer part of the migrations). 

Another good catch! Although I do agree with you on its importance, I preferred to fix it right now.

> 1. if parameters are invalid for the deployment of a collector the revert reason is not shown, instead ...

I'm not sure what we could do to validate all the input params. We may validate the token address (maybe) and the partner shares as well, but I'm not sure how to validate the other addresses; let's discuss this.

mortelli (commented):
token address param works OK :+1: 

i had some issues with the withdraw scripts but i'd rather discuss them through a call tomorrow

thanks for the fixes!

mortelli:
got it to work, my only note is that we could add some parametrization here and there

PR #22, opened 2022-02-23
PP-94: Revenue Sharing Model

mortelli:
# Revenue Sharing Model

This PR is part of the implementation of the Revenue Sharing model feature.

## Business Logic overview

The Revenue Sharing Model is a way to relay transactions so that the resulting relay fees are shared among multiple partners. 

A new type of contract (called "Collector") can now be used in relay requests, and will act as the receiver of its related relay fees. The usual receivers of these fees are the relay workers, and this will continue to be the case if a Collector is not used. This is, in fact, the way to opt out of the Revenue Sharing feature. 

At a later (and arbitrary) point in time, collected fees can be withdrawn from the contract and distributed between the different partners involved (also called "beneficiaries") which are addresses written in the contract, each with their own percentage of shares.  

For more in-depth information please see [the devportal Revenue Sharing documentation](https://github.com/mortelli/devportal/pull/20). 

## Technical overview

1. The Collector smart contract has been added. 
    1. It has a public token and owner, as well as a private set of revenue sharing partners. 
    2. The revenue sharing update, withdraw & ownership transfer functions can only be called by the owner.
2. The `collectorContract` field has been added to the `ForwardRequest` interface.
3. The Smart Wallet and Custom Smart Wallet now pay relay fees to the Collector address if specified, and to the relay worker in other cases.
4. Tasks for creating a safe & deploying a collector have been added.
    1. The `create-safe` script is useful in the cases in which a Collector needs to be deployed using a [multisig address](https://github.com/gnosis/safe-contracts) as its owner.
    2. Having a specific task for deploying a collector comes in handy because this process is outside of the RIF Relay migrations and therefore not automatic. Documentation on how to deploy a Collector using this script has been added to the Readme file.
5. General improvements have been applied to other tasks/scripts.

### Other Revenue Sharing PRs
1. `rif-relay-common`: [PR #10 (Revenue Sharing Model)](https://github.com/infuy/rif-relay-common/pull/10)
2. `rif-relay-client`: [PR #12 (Revenue Sharing Model)](https://github.com/infuy/rif-relay-client/pull/12)
3. `rif-relay-server`: [PR #16 (Revenue Sharing Model)](https://github.com/infuy/rif-relay-server/pull/16)
4. `rif-relay`: [PP #194 (Revenue Sharing Model)](https://github.com/rsksmart/rif-relay/pull/194)
5. `relaying-services-sdk`: [PR #14 (Revenue Sharing Model)](https://github.com/infuy/relaying-services-sdk/pull/14)
6. `relaying-servives-sdk-dapp`: [PP #13 (Revenue Sharing Model)](https://github.com/antomor/relaying-services-sdk-dapp/pull/13)
7. `devportal`: [Revenue Sharing documentation](https://github.com/mortelli/devportal/pull/20)

### Revenue Sharing Jira issues
See [here](https://rsklabs.atlassian.net/issues/?jql=labels%20%3D%20%22revenue-sharing%22).

### Other 
[RIF Relay - Revenue Sharing implementation](https://docs.google.com/document/d/1lmh78Bv7my5z6gkLEHuSqAj4NIQxZw6JwwfFT6zd9Dw)


contract-addresses.json
```diff
@@ -1,16 +1,30 @@
 {
+    "31": {
```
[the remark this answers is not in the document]
mortelli: i've manually checked this file against `master` and there are no changes, not sure why it's showing up in the diff. even if we have squashed before, showing changes here isn't correct. 

`git diff PP-94/revenue-sharing-model master -- contract-addresses.json` is showing no diffs

truffle.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
     networks: {
         development: {
             verbose: process.env.VERBOSE,
-            host: '127.0.0.1',
+            host: '172.18.0.2',
```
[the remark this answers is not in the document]
mortelli: done

revenue-sharing-addresses.json
```diff
@@ -0,0 +1,23 @@
+{
```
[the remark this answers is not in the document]
mortelli: done

mortelli:
requesting review from @antomor

contracts/Collector.sol
```diff
@@ -0,0 +1,76 @@
...
+pragma experimental ABIEncoderV2;
+
+import "./interfaces/ICollector.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/math/SafeMath.sol";
+
+contract Collector is ICollector{
+    IERC20 public token;
+    address public owner;
+    RevenuePartner[] private partners;
```
antomor, line 12: As we already talked about, we could expose the partner addresses only. Not required for the current purpose, but do we need to keep track of it?

antomor, line 12: As we already talked about, we could expose the partner addresses only. Not required for the current purpose, but do we need to keep track of it?

contracts/Collector.sol
```diff
@@ -0,0 +1,76 @@
...
+        for(uint i = 0; i < partners.length; i++)
+            token.transfer(partners[i].beneficiary, SafeMath.div(SafeMath.mul(balance, partners[i].share), 100));
+    }
+
+    function transferOwnership(address _owner)
+    external 
+    override
+    onlyOwner()
+    {
+        owner = _owner;
```
antomor, line 59: I just checked the [Openzeppelin implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4a9cc8b4918ef3736229a5cc5a310bdc17bf759f/contracts/access/Ownable.sol#L62) and I think we are missing the check of the address received against `address(0)`

contract-addresses.json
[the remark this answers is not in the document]
antomor: Yes, I think that maybe we changed it in other PRs, but I agree that it shouldn't appear now

antomor (changes requested):
Thanks for opening this PR 🙏🏼 
I've left a couple of doubts we may want to tackle before merging it.

contracts/Collector.sol
mortelli: i have added this to the pending issues list so we can discuss it, i haven't created a JIRA for it since i don't know if we're actually going to move forwards with it

contracts/Collector.sol
```diff
@@ -0,0 +1,76 @@
...
+            partners.push(_partners[i]);
+    }
+
+    function withdraw() 
+    external 
+    override
+    onlyOwner()
+    {
+        uint balance = token.balanceOf(address(this));
+        require(balance > 0, "no revenue to share");
```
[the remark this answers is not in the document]
antomor: Fine, thank you, I agree with you that if it's easy for the user to check the collector balance, we don't need to change anything here 👍🏼

contracts/Collector.sol
antomor: I know that the owner can specify any address there and we cannot check the validity of the address, but if they specify `address(0)` we are sure the collector will be unusable at all

antomor (approved):
Thanks for applying the changes!

contracts/Collector.sol
mortelli: makes sense :+1:

PR #23, opened 2022-03-01
PP-131:Parametrize the withdraw and create safe scripts

antomor:
## What

- Change the `create-safe` script to receive owners as an input parameter;
- Change the `withdraw` script to receive the `tokenAddress` as an input parameter;
- Change the `withdraw` script to allow its execution whether the collector owner is a multisig account or an EOA;
- Remove the `send-multisig` script since it's unused.

## Why

- The addresses used to create a safe account were hard-coded;
- The token used in the withdraw script was the `TestToken`, so it wasn't ERC-20 compliant;
- The `withdraw` script doesn't support collector Contract where the owner isn't a multisig account.


mortelli (commented):
thanks for this PR :raised_hands: 

i've left some low-level comments below, but here are my general notes.

these were my manual verification steps:
1. create a safe specifying 2 owners ("safe A") :heavy_check_mark: 
2. create a safe without specifying owners ("safe B") :heavy_check_mark: 
3. deploy 3 collectors: one for safe A, one for safe B, and a final one for an EOA owner :heavy_check_mark: ("collector C")
4. mint tokens to all 3 collectors :heavy_check_mark: 
5. call the withdraw script for collector A :red_circle: 
6. call the withdraw script for collector B :red_circle: 
7. call the withdraw script for collector C :heavy_check_mark: 

i noticed that the withdraw calls for collectors A & B were failing due to missing signatures. after looking at the code, i realized that the owners for each of these safes were being taken from the revenue sharing addresses file.

however, in my case, this file indicated partners for the collectors, which were different from the safe owners. i think in most cases they will match, but we might want to find a way to support this other scenario.

i know it might be verbose, but maybe passing the owners as parameters instead of reading them from the file?

if opting for this option, i would also eliminate default values for both the collector and the token fields, since it could happen that after deploying multiple collectors the file in question would not have correct values for them.

as a side note: just to check, i entered the safe owners as "partners" in this file, just to get their signatures for the withdraw operation. while the call does not fail, the collector balance does not change either (looks like no funds are actually being distributed in this case). however, i am willing to take a look at this again after the previous point is tackled.

finally, a small detail: 
> The token used in the withdraw script was the TestToken, so it wasn't ERC-20 compliant;

are you sure about this? i've checked the code and it says `is ERC20`.

Readme.md
```diff
@@ -48,6 +48,33 @@ The contracts can be deployed in the following way:
...
+To call the `withdraw` method available on the Collector contract, the repo provides the script `/tasks/withdraw.js`. It allows the user to call the `withdraw` whether the contract owner is a Multisig account or an EOA.
+
+```bash
+npx truffle --network regtest exec tasks/withdraw.js --safeAddress='<0xsafe_address>' --collectorAddress='<0xcollector_address>' --tokenAddress='<0xtoken_address>'
+```
+
+Parameters:
+* `safeAddress`: Address of the Safe account set as owner of the Collector. It must be used only when then collector owner is a multisig account;
+* `collectorAddress`: Address of the Collector deployed; if not specified, the `collectorContract` field of the `revenue-sharing-addresses.json` file is used.
+* `tokenAddress`: Address of the ERC-20 token used by the Collector; if not specified, the `collectorToken` field of the `revenue-sharing-addresses.json` file is used.
```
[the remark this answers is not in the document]
antomor: Good point, thanks! I think it's even better to move it to the `System usage` section

tasks/create-safe.js
```diff
@@ -82,9 +87,11 @@ async function deploySafe(
             ZERO_ADDRESS
         ]
     );
+    // TODO: Value extracted from regtest, likely we need to adapt it for testnet/mainnet
+    const createTxGasLimit = '1081903';
```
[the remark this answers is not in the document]
antomor: Yes, probably we should

antomor:
Thanks for the review, I think there was room for improvement.

> i know it might be verbose, but maybe passing the owners as parameters instead of reading them from the file?
> 
> if opting for this option, i would also eliminate default values for both the collector and the token fields, since it could happen that after deploying multiple collectors the file in question would not have correct values for them.

I like your suggestions, so although the script may become verbose, I applied both, because they make the script easier to understand and to run.

> as a side note: just to check, i entered the safe owners as "partners" in this file, just to get their signatures for the withdraw operation. while the call does not fail, the collector balance does not change either (looks like no funds are actually being distributed in this case). however, i am willing to take a look at this again after the previous point is tackled.
 
Just re-tested everything and it seems that after executing the script the token is distributed properly. If you see it again, maybe we can investigate further on that.

> > The token used in the withdraw script was the TestToken, so it wasn't ERC-20 compliant;
>
> are you sure about this? i've checked the code and it says `is ERC20`.

I meant that the script didn't support all erc20 token, but just TestToken, so in this sense, the script wasn't ERC-20 compliant, but it supported just this specific implementation of an ERC-20 token.

tasks/withdraw.js
```diff
@@ -63,32 +95,63 @@ module.exports = async (callback) => {
...
+                gasLimit: safeTxGasLimit
+            }
+        );
+        withdrawTxHash = safeTxResponse.hash;
+    } else {
+        // TODO: We saw that '137410' is required in regtest, likely this value is going to change on testnet/mainnet
+        const gasRequired = '140000';
+        const tx = await web3.eth.sendTransaction({
+            ...withdrawTx,
+            from: '0xCD2a3d9F938E13CD947Ec05AbC7FE734Df8DD826',
```
mortelli, line 145: we need to change this one. we might need a new parameter, since the transaction must come from the collector owner

mortelli (commented):
thanks for all the corrections :+1: 

the previous cases are working OK now, but i found an error in the last scenario (withdrawal with no safe).

i am ready to approve as soon as this is solved :slightly_smiling_face:

tasks/withdraw.js
antomor: My fault, I'm sorry, you're right, and I'd propose that we use directly the `owner` field in the collector instead of using an additional parameter ;-)

mortelli: superb! thanks

mortelli (approved):
thanks for the last correction, my turns turned out a-ok :ok_hand:

PR #24, opened 2022-04-08
PP-131/lazy-load-revenue-sharing-addresses

antomor:
## What

- Import the `revenue-sharing-addresses.json` file when required to avoid errors when executing `create-safe`.

## Why

- the file `revenue-sharing-addresses.json` is created when the collector is deployed, but we need to execute the `create-safe` script before and it requires `utils.js`. ATM `utils.js` requires the `revenue-sharing-addresses.json` file at the beginning, so if the file is missing, it raises an error.

### DIFF
--- a/tasks/utils.js
+++ b/tasks/utils.js
@@ -1,5 +1,4 @@
 const safeCoreSdk = require('@gnosis.pm/safe-core-sdk');
-const RevenueSharingAddresses = require('../revenue-sharing-addresses.json');
 const { Web3Adapter } = safeCoreSdk;
 
 const getTestTokenInstance = async (artifacts) => {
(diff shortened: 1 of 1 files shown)


tasks/utils.js
```diff
@@ -79,9 +78,21 @@ const contractNetworks = {
 
 const getPartnerKey = (index) => `partner${index}`;
 
+const defaultRevenueSharingAddressesPath = '../revenue-sharing-addresses.json';
+let revenueSharingAddresses = undefined;
+const getRevenueSharingAddressesFromFile = (
+    filename = defaultRevenueSharingAddressesPath
+) => {
+    if (revenueSharingAddresses === undefined) {
```
ironFe93, line 86: I think this line could be like this:
`if(!revenueSharingAddresses) {`

tasks/utils.js
```diff
@@ -79,9 +78,21 @@ const contractNetworks = {
...
+        revenueSharingAddresses = require(filename);
+    }
+    return revenueSharingAddresses;
```
ironFe93, line 89: What happens if revenueSharingAddresses is still undefined? Should an error be thrown?

antomor: Yes, and that's what happens if the file is missing. This should be the default behaviour of `require` in nodejs

tasks/utils.js
antomor: Although the result may be the same, I usually prefer to be explicit and check for the type as well.
In this specific case, we set the initial value to `undefined` and this is the only case in which I would read it from the file.

ironFe93 (approved):
LGTM

PR #26, opened 2022-05-19
Pp 148/setup contract tests

jd-iov:
## What

- Dependencies needed to execute testing suites or single tests for the project from folder **test**

## Why

- Tests related to smart contracts will be added to this repo and these configurations were needed in order to initiate the testing process.

## Refs

- [Jira ticket](https://rsklabs.atlassian.net/browse/PP-148)


antomor (approved):
Thank you, LGTM!
I would add a sample test (like the one you showed to me) to be sure we are aligned on the fact that we want to write tests using ts and not js.

antomor (approved):
LGTM!

PR #27, opened 2022-05-22
PP-147: feature (devops): add ci steps for rif-relay-contracts

david-iov:


antomor (changes requested):
@david-iov Thank you for this change.
If you don't mind I'd like to include also the test step, so I'd wait for [this PR](https://github.com/rsksmart/rif-relay-contracts/pull/26) to be merged and then I'll ask you to update this accordingly. WDYT?

.github/workflows/main.yml
```diff
@@ -0,0 +1,29 @@
...
+                  node-version: '16.x'
+
+            - name: Install dependencies
+              run: npm install
+
+            - name: Check Linter bugs for Typescript
+              run: npm run lint:ts
+
+            - name: Fix Linter bugs for Typescript
+              run: npm run lint:ts:fix
```
antomor, line 20: I think that this task will never be executed because it's necessary when the previous one fails, so it's intended to be used locally by devs.

.github/workflows/main.yml
```diff
@@ -0,0 +1,29 @@
...
+
+            - name: See bugs on solidity
+              run: npm run lint:sol
+
+            - name: Check Codestyles errors
+              run: npm run prettier
+
+            - name: Fix Codestyles bugs
+              run: npm run prettier:fix
```
antomor, line 29: I think that this task will never be executed because it's necessary when the previous one fails, so it's intended to be used locally by devs.

antomor (changes requested):
I'd remove the *:fix tasks because they are executed when the check tasks fail

antomor (approved):
@david-iov Thank you LVGTM!

PR #32, opened 2022-06-08
patch: removes dist and old scripts

jurajpiar:
## What

- removes dist folder and old scripts
- adds build to prepare script

## Why

- Reducing inconsistencies in the rif-relay projects

## Refs

- https://rsklabs.atlassian.net/browse/PP-166
- https://rsklabs.atlassian.net/browse/PP-74


antomor (changes requested):
Thank you for these changes. If we don't use it anymore, could we please remove `pinst` from the package.json dependencies?
Also, we need to update the `Readme.md` file to reflect these changes; I would update the [installation](https://github.com/rsksmart/rif-relay-contracts#as-a-dependency) section as well.

antomor (approved):
LGTM!

PR #33, opened 2022-06-15
PP-163: Smartwallet tests

jd-iov:
## What

- Tests for CustomSmartWallet and SmartWallet contracts 

## Why

- Increase test coverage over these contracts by adding these unit tests

## Refs

- Please refer to the [ticket](https://rsklabs.atlassian.net/browse/PP-163)


test/smartwallet/customSmartWallet.test.ts
```diff
@@ -0,0 +1,752 @@
...
+            gas: '1000000',
+            nonce: '0',
+            data: '0x',
+            tokenContract: constants.ZERO_ADDRESS,
+            tokenAmount: '1',
+            tokenGas: '50000'
+        },
+        relayData: {
+            gasPrice: '1',
+            domainSeparator: '0x',
```
AndresQuijano, line 91: The domainSeparator is no longer required, according to this task https://rsklabs.atlassian.net/browse/PP-137. We have to coordinate the merge. See [this](https://github.com/rsksmart/rif-relay/pull/201/files#diff-3f69b22eb4a55fe2bec311acd54690e7a0611ea34a6f2c118717bfacbaa0579aL185)

franciscotobar (commented):
The TestUtils.ts is going to be individual for each project?

test/utils/TestUtils.ts
```diff
@@ -0,0 +1,81 @@
+export function bytes32(n: number): string {
+    return '0x' + n.toString().repeat(64).slice(0, 64);
+}
```
jurajpiar, line 3: replace with `web3.utils.toHex`

antomor:
> The TestUtils.ts is going to be individual for each project?

It depends on its usage. If we notice we need it in other projects we could decide to export it from here or to create a utility library for testing.

test/smartwallet/CustomSmartWallet.test.ts
```diff
@@ -0,0 +1,1197 @@
...
+                        '10',
+                        '400000',
+                        '0x'
+                    )
+                );
+
+                assert.isTrue(await smartWallet.isInitialized());
+            });
+
+            it('Should verify method initialize successfully with worker address as null parameter', async () => {
```
AndresQuijano, line 275: Is this test description correct? The assert says isFalse.

test/utils/TestUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export type IntString = string;
```
jurajpiar, line 5: If you want to enforce integer in a string then perhaps you would like to use
``  export type IntString = `${number}` ``

test/utils/TestUtils.ts
antomor: I don't think they do exactly the same. This function actually repeat the same number 64 times and it returns it with `0x` as prefix, while `web3.utils.toHex` actually returns its hexadecimal value. Am I missing anything?

package.json
```diff
@@ -32,19 +32,20 @@
...
-    "eth-sig-util": "2.5.2",
-    "ethereumjs-tx": "~2.1.2",
-    "ethereumjs-util": "~6.2.1",
     "@openzeppelin/contracts": "~3.2.0",
     "@truffle/contract": "4.2.23",
     "@truffle/contract-schema": "3.3.2",
     "@truffle/hdwallet-provider": "1.6.0",
+    "eth-sig-util": "2.5.2",
+    "ethereumjs-tx": "~2.1.2",
+    "ethereumjs-util": "~6.2.1",
```
jurajpiar, line 41: I have a feeling we don't actually need this, but that's out of scope, I guess.

package.json
```diff
@@ -32,19 +32,20 @@
...
   },
   "dependencies": {
-    "eth-sig-util": "2.5.2",
-    "ethereumjs-tx": "~2.1.2",
-    "ethereumjs-util": "~6.2.1",
     "@openzeppelin/contracts": "~3.2.0",
     "@truffle/contract": "4.2.23",
     "@truffle/contract-schema": "3.3.2",
     "@truffle/hdwallet-provider": "1.6.0",
+    "eth-sig-util": "2.5.2",
```
jurajpiar, line 39: deprecated!!

package.json
```diff
@@ -32,19 +32,20 @@
...
+    "ethereumjs-tx": "~2.1.2",
```
jurajpiar, line 40: archived and replaced by `@ethereumjs/tx`

test/utils/TestUtils.ts
jurajpiar: well, that's the bit I don't understand. Why does it need to do that? In any case, the intent should be captured in the name of the function.

test/smartwallet/CustomSmartWallet.test.ts
antomor: Good catch, thank you for highlight this. In my opinion, I think that this test could be removed, because we set the address with a null value where the null value shouldn't be allowed; indeed it's an error for typescript, but it looks like truffle doesn't perform the ts check before running the tests. Maybe we could try to configure truffle to check the types when running the tests.

test/utils/TestUtils.ts
antomor: Yes, I've tried to rename it to make it clear what it does

test/utils/TestUtils.ts
antomor: It's not used this type, so I removed it.

package.json
antomor: It's used in `types/EIP712/TypedRequestData.ts#L10` right now. I think we could review them in the related task.

package.json
antomor: I think we could review this and the other dependencies in the related task

package.json
antomor: I think we could review this and the other dependencies in the related task

test/smartwallet/CustomSmartWallet.test.ts
antomor: For now, I've updated the description to reflect what it actually does

test/utils.ts
```diff
@@ -1,11 +1,9 @@
 import RelayHubConfiguration from '../types/RelayHubConfiguration';
 
-export function bytes32(n: number): string {
+export function createBytes32FromNumber(n: number): string {
     return '0x' + n.toString().repeat(64).slice(0, 64);
 }
```
jurajpiar, line 5: Sorry if it's knit-picking. I think it has to be clear though, that the generated string is not a bytes32 representation of the given number.
```suggestion
export function generateBytes32(seed: number): string {
    return '0x' + seed.toString().repeat(64).slice(0, 64);
}
```

test/utils/TestUtils.ts
jurajpiar: I've made a change suggestion in a different comment.

jurajpiar (approved):
LG

AndresQuijano (approved):
LGTM

test/smartwallet/customSmartWallet.test.ts
antomor: Yes, we rebased the branch

PR #35, opened 2022-06-20
PP-165: Update workflow to use official rskj node image

david-iov:


antomor (commented):
@david-iov Thank you for this PR. The changes look ok. My only concern regards the usage of the latest version. IMHO we should use the stick with version 3.x.y, to avoid errors when a major is released. The point is that at the moment we don't have any tag that would help in this. I'm waiting for a response from Lucas in the related thread on slack.

david-iov:
> @david-iov Thank you for this PR. The changes look ok. My only concern regards the usage of the latest version. IMHO we should use the stick with version 3.x.y, to avoid errors when a major is released. The point is that at the moment we don't have any tag that would help in this. I'm waiting for a response from Lucas in the related thread on slack.

@antomor I've made the change to use the latest stable version tag

antomor:
@david-iov thank you for the updates, but I wouldn't use a specific version otherwise we should keep updating it each time a new version is available.
If available I'd use a tag that uses the latest stable version given a major number, so for instance `IRIS-3` would point to any 3.x.y stable version and we would need to update just after a major release.

antomor (approved):
LGTM!

PR #41, opened 2022-07-11
PP-127: collector contract optional param

ironFe93:
## What

Overload execute function in Smart Wallet Contract

## Why

Collector Contract address should be optional

## Refs

PP-94/revenue-sharing-model


contracts/RelayHub.sol
```diff
@@ -224,9 +225,14 @@ contract RelayHub is IRelayHub {
...
-        //use succ as relay call success variable
-        (forwarderSuccess, destinationCallSuccess, relayedCallReturnValue) = Eip712Library
-            .execute(relayRequest, signature);
+        if(collectorContract == address(0)) {
+            (forwarderSuccess, destinationCallSuccess, relayedCallReturnValue) = Eip712Library
+                .execute(relayRequest, signature);
+        } else {
+            (forwarderSuccess, destinationCallSuccess, relayedCallReturnValue) = Eip712Library
+                .execute(relayRequest, signature, collectorContract);
+        }
```
jurajpiar, line 234: unnecessary check, as `execute(relayRequest, signature)` will call the method with `address(0)`, which is the collectorContract in this case anyway.

contracts/RelayHub.sol
```diff
@@ -195,7 +195,8 @@ contract RelayHub is IRelayHub {
 
     function relayCall(
         EnvelopingTypes.RelayRequest calldata relayRequest,
-        bytes calldata signature
+        bytes calldata signature,
```
jurajpiar, line 198: let's set a precedence here and let's use explicit types.
```suggestion
        bytes32 calldata signature,
```

contracts/RelayHub.sol
jurajpiar, line 234: unnecessary check, as `execute(relayRequest, signature)` will call the method with `address(0)`.
```suggestion  
     (forwarderSuccess, destinationCallSuccess, relayedCallReturnValue) = Eip712Library
                .execute(relayRequest, signature, collectorContract);
```

contracts/RelayHub.sol
ironFe93: done

contracts/RelayHub.sol
ironFe93: done

contracts/utils/Eip712Library.sol
```diff
@@ -30,11 +30,23 @@ library Eip712Library {
...
+                    abi.encodeWithSelector(IForwarder.execute.selector, relayRequest.relayData.domainSeparator,
+                    hashRelayData(relayRequest.relayData), relayRequest.request, signature
+                    ));
+            } else {
+                /* solhint-disable-next-line avoid-low-level-calls */
+                (forwarderSuccess, ret) = relayRequest.relayData.callForwarder.call(
+                    abi.encodeWithSelector(IForwarder.execute.selector, relayRequest.relayData.domainSeparator,
+                    hashRelayData(relayRequest.relayData), relayRequest.request, signature, collectorContract
+                    ));
+            }
```
jurajpiar, line 49: can we not just send the contract address anyway?

contracts/RelayHub.sol
antomor: @jurajpiar @ironFe93  Are we sure that it doesn't break anything? In my understanding they are two different types, and I don't remember the size of the signature, so I'm not sure if it fits in a bytes32.

contracts/utils/Eip712Library.sol
antomor: What do you mean exactly?

contracts/smartwallet/SmartWallet.sol
```diff
@@ -154,7 +155,22 @@ contract SmartWallet is IForwarder {
   
     }
 
-   
+   function execute(
```
antomor, line 158: I think we should add this method in the interface also

contracts/utils/Eip712Library.sol
```diff
@@ -30,11 +30,23 @@ library Eip712Library {
...
-                hashRelayData(relayRequest.relayData), relayRequest.request, signature
-                ));
+                (forwarderSuccess, ret) = relayRequest.relayData.callForwarder.call(
+                    abi.encodeWithSelector(IForwarder.execute.selector, relayRequest.relayData.domainSeparator,
+                    hashRelayData(relayRequest.relayData), relayRequest.request, signature
+                    ));
+            } else {
+                /* solhint-disable-next-line avoid-low-level-calls */
+                (forwarderSuccess, ret) = relayRequest.relayData.callForwarder.call(
+                    abi.encodeWithSelector(IForwarder.execute.selector, relayRequest.relayData.domainSeparator,
```
antomor, line 46: If we don't add the method in the interface, we won't be able to find it in the `IForwarder`

ironFe93:
closed as this proposal will not go through.

contracts/RelayHub.sol
jurajpiar: bytes == bytes32... it's just better to be expressive about size

jurajpiar: actually, after re-reading the docs, I see this is not the case :/
`bytes` type is tightly packed (so it may have a smaller footprint), whereas `bytesN` will aways take the N-bytes of space.

PR #42, opened 2022-07-12
PP-271: Include contracts/test to allow the test execution

antomor:
## What

- publish the `contracts/test` folder

## Why

- the `contracts/test` folder includes some contracts useful for testing purposes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,7 @@
     "ignore": [
       "!dist",
       "!build",
-      "!contracts/test",
       "src",
-      "test",
       ".github"
     ]
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -27,7 +27,6 @@
       "!dist",
       "!build",
       "src",
-      "test",
```
jurajpiar, line 30: ah nice one. Maybe it just would be better to use `"!test"` to make it explicit.

antomor: Good point, thank you!

jurajpiar (approved):
LG

ironFe93 (approved):

PR #44, opened 2022-07-21
PP-130/no remainders

ironFe93:
## What

Set up functionality to enable updating partners and shares after withdrawing any amount of tokens 

## Why

Currently one can only update partners when token balance = 0, but due to how integer division works in solidity, it is very likely that there will be a remainder after distributing all fees to the partners. a token amount of 999 that needs to be distrubuted to 4 partners would leave a remainder of balance = 3 after withdrawal. 
## Refs

- PP-130, PP-94


contracts/Collector.sol
```diff
@@ -3,17 +3,17 @@ pragma solidity ^0.6.12;
 pragma experimental ABIEncoderV2;
 
 import "./interfaces/ICollector.sol";
-import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/ERC20Burnable.sol";
 import "@openzeppelin/contracts/math/SafeMath.sol";
 
 contract Collector is ICollector{
-    IERC20 public token;
+    ERC20Burnable public token;
```
antomor, line 10: Although it makes sense, I'm not sure we can use this interface. It extends the [EIP-20 standard](https://eips.ethereum.org/EIPS/eip-20) with additional functions that may be not implemented by the token.

contracts/Collector.sol
```diff
@@ -56,6 +56,11 @@ contract Collector is ICollector{
 
         for(uint i = 0; i < partners.length; i++)
             token.transfer(partners[i].beneficiary, SafeMath.div(SafeMath.mul(balance, partners[i].share), 100));
+        
+        balance = token.balanceOf(address(this));
```
antomor, line 60: Why are we burning the token here? I think a lazy approach would be better because if the same partners continue to operate, the reminder may be distributed within the next `withdraw` call. In this way, we are always burning the reminder.

test/Collector.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        return { collector, testToken, partners, owner };
+    }
+
+    describe('Deployment', function () {
+        it('Should deploy with owner, token and revenue partners', async function () {
+            const { collector } = await deployCollector();
+            expect(collector.address).to.be.properAddress;
+        });
+
+        it('Should not let deploy with invalid shares', async function () {
```
antomor, line 52: Thank you for adding these tests. What do you think if we cover more scenarios? e.g.: negative shares, no shares, etc...

test/Collector.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+    describe('updateShares', function () {
+        it('Should update shares and partners when token balance is zero', async function () {
+            const { collector, partners } = await deployCollector();
+            await collector.updateShares(partners);
+            expect('updateShares').to.be.calledOnContract(collector);
+        });
+
+        it('Should update shares and partners after withdrawing irregular amount', async function () {
```
antomor, line 85: what do you mean by "irregular amount"? Maybe you mean with an amount that generates reminder right?

package.json
```diff
@@ -18,7 +19,8 @@
     "allowedTokens": "./scripts/allowedTokens",
     "prettier:fix": "npx prettier --write --ignore-unknown .",
     "prettier": "npx prettier --check .",
-    "test": "truffle test --network regtest"
+    "test": "truffle test --network regtest",
+    "test-waffle": "NODE_ENV=test mocha"
```
antomor, line 23: what's the difference between `test` and `test-waffle`. I think we should update the readme to include it

package.json
```diff
@@ -6,6 +6,7 @@
   "private": false,
   "scripts": {
     "dist": "scripts/dist",
+    "build": "waffle",
```
antomor, line 9: what's the difference between `build` and `dist`? We should document it in the readme

antomor (changes requested):
@ironFe93 I've left some comments. Maybe we need to understand the way we want to implement to burn the token

test/Collector.test.ts
ironFe93: Yes that's what I meant

contracts/Collector.sol
ironFe93: This is the key part. If not all tokens implement the burnable method then this proposal falls apart and I will think about something different.

package.json
ironFe93: went with just dist, since it is required in the git pipeline

package.json
ironFe93: keeping test:truffle for contract migration. Maybe we should change this though...

test/Collector.test.ts
ironFe93: changed to "remainder amount"

contracts/Collector.sol
ironFe93: completely revamped.

contracts/Collector.sol
```diff
@@ -73,12 +83,23 @@ contract Collector is ICollector{
...
     }
 
     modifier onlyOwner(){
         require(msg.sender == owner, "can only call from owner");
         _;
     }
+
+    modifier noRemainders(){
+        uint balance = token.balanceOf(address(this));
+        require(balance < partners.length, "balance is not a remainder");
```
AndresQuijano, line 96: I suggest a more clear message here. Something like "the balance is greater than a remainder" or "there are tokens to share" or something else.

AndresQuijano (changes requested):
The "What" of this PR says something related to testing but the core is more related to the remainder stuff. I suggest to modify it to reflect the intention of this task.

scripts/prepare
```diff
@@ -3,5 +3,6 @@
 patch-package
 
 npm run dist
+npm run build
```
AndresQuijano, line 6: This script is missing in package.json and when I execute npm run prepare, it throws an npm error.

test/Collector.test.ts
```diff
@@ -0,0 +1,185 @@
...
+use(solidity);
+
+const buildPartners = (
+    wallets: Wallet[],
+    share1?: number,
+    share2?: number,
+    share3?: number,
+    share4?: number
+) => {
+    const [owner, remainder, partner1, partner2, partner3, partner4] = wallets;
```
AndresQuijano, line 22: I wonder if this line can be replaced by:
const [, , partner1, partner2, partner3, partner4] = wallets;

test/Collector.test.ts
```diff
@@ -0,0 +1,185 @@
...
+        testToken.address,
+        partners,
+        remainderWallet.address
+    ]);
+
+    return { collector };
+}
+describe('Collector', function () {
+    async function deployTokenFixture(wallets) {
+        const testToken = await deployContract(wallets[0], TestToken);
```
AndresQuijano, line 62: It's not clear what's in the 0 position of this array. Perhaps using a constant could be useful. Something like wallets[TOKEN_POSITION] or wallets[OWNER_POSITION] and at top of the file or in a separate one: OWNER_POSITION=0;
Otherwise the positions of the array can be documented on a comment. 
See [https://refactoring.guru/replace-magic-number-with-symbolic-constant](url) and [https://stackoverflow.com/questions/2089429/magic-numbers-in-arrays-c](url)

test/Collector.test.ts
```diff
@@ -0,0 +1,185 @@
...
+        ownerWallet.address,
+        testToken.address,
+        partners,
+        remainderWallet.address
+    ]);
+
+    return { collector };
+}
+describe('Collector', function () {
+    async function deployTokenFixture(wallets) {
```
AndresQuijano, line 61: Is it possible to type the wallets parameter?

contracts/test/tokens/BurnableTestToken.sol
```diff
@@ -0,0 +1,15 @@
...
+contract BurnableTestToken is ERC20("Test Token", "TKN") {
+    function mint(uint amount, address to) public {
+        _mint(msg.sender, amount);
+        _transfer(msg.sender, to, amount);
+    }
+
+    function burn(uint amount) public {
+        _burn(msg.sender, amount);
+    }
+}
```
AndresQuijano, line 15: Please tell me where are you using this. I couldn't see it in use.

contracts/Collector.sol
```diff
@@ -73,12 +83,23 @@ contract Collector is ICollector{
...
+        if(balance == 0) {
+            return;
+        }
+        
+        token.transfer(remainderAddress, balance);
+            balance = token.balanceOf(address(this));
```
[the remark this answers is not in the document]
ironFe93: the purpose of this modifier is to clear remainders left in the contract. Maybe the modifier functions should be called "clearRemainders" to be more explicit.

contracts/Collector.sol
```diff
@@ -73,12 +83,23 @@ contract Collector is ICollector{
...
+        _;
+    }
```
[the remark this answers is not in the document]
ironFe93: Good find, agreed

contracts/Collector.sol
```diff
@@ -29,16 +32,23 @@ contract Collector is ICollector{
...
         delete partners;
         
         for (uint i = 0; i < _partners.length; i++)
             partners.push(_partners[i]);
     }
 
+    function updateRemainderAddress(address _remainderAddress) 
+    external
+    onlyOwner()
+    noRemainders()
```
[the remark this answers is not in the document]
ironFe93: The idea is that the remainder address owns the right to any remainder balance that still remains in the contract before calling this function. This enforces that the remainderAddress can only be changed after a withdrawal. It is what made most sense to me but it's open to discuss.

contracts/Collector.sol
ironFe93: We are contrained by a linter error "error message for require is too long". How about "balance > remainder"?

contracts/Collector.sol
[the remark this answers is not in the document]
AndresQuijano: No, I mean, the line 102 is updating the balance variable and I don't understand its purpose.

test/Collector.test.ts
ironFe93: agreed

test/Collector.test.ts
ironFe93: Agree on being more explicit. Will use constants instead.

test/Collector.test.ts
ironFe93: yes, done.

contracts/test/tokens/BurnableTestToken.sol
ironFe93: You are right, It's a leftover and I will remove it.

scripts/prepare
ironFe93: it's a leftover, I will remove it, thanks.

contracts/Collector.sol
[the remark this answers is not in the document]
antomor: At the moment it's the owner of the contracts that allows the function to be called. Probably now that we introduced the `remainderAddress` it doesn't make sense this check either here or in the `updateShares` method. WDYT? After all, it's the owner that decide what to do, and with the multisig the partner should agree when changing the `remainderAddress`.

contracts/Collector.sol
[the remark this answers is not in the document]
ironFe93: it shouldn’t be there, I'm removing it

contracts/Collector.sol
```diff
@@ -73,12 +83,23 @@ contract Collector is ICollector{
...
     modifier onlyOwner(){
         require(msg.sender == owner, "can only call from owner");
         _;
     }
+
+    modifier noRemainders(){
+        uint balance = token.balanceOf(address(this));
+        require(balance < partners.length, "balance is not a remainder");
+        if(balance == 0) {
+            return;
```
antomor, line 98: I'm not sure about this line. We'll probably still need to use `_`. We need to check because I don't remember exactly.

contracts/Collector.sol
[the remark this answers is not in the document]
antomor: I think the indentation is wrong here.

contracts/Collector.sol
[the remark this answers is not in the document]
antomor: After checking the code I think we may need to split the operations performed in the `noRemainders` modifier. Because we may probably want to send the remainder in some cases, and not sure if we want to check that there's no remainder

contracts/Collector.sol
```diff
@@ -9,18 +9,21 @@ import "@openzeppelin/contracts/math/SafeMath.sol";
...
     address public owner;
+    address private remainderAddress;
     RevenuePartner[] private partners;
 
     constructor(
         address _owner,
         IERC20 _token,
-        RevenuePartner[] memory _partners
+        RevenuePartner[] memory _partners,
+        address _remainderAddress
```
antomor, line 19: Do you see any addition in providing a method to update both partners and reminderAddress in one single transaction?

package.json
```diff
@@ -18,7 +18,8 @@
     "allowedTokens": "./scripts/allowedTokens",
     "prettier:fix": "npx prettier --write --ignore-unknown .",
     "prettier": "npx prettier --check .",
-    "test": "truffle test --network regtest"
+    "test:truffle": "truffle test --network regtest",
+    "test": "NODE_ENV=test mocha --timeout 10000"
```
antomor, line 22: what this timeout is used for? Is it necessary for all the tests, or do we need it in some specific tests?

antomor (changes requested):

package.json
ironFe93: It is needed specially for unit tests that test the contracts deployment, however I have observed that other tests can reach timeout as well.

contracts/Collector.sol
ironFe93: Maybe we can do this in a separate task? I do think it makes sense to have this functionality available (through an overloaded function) but I faced an issue while testing it.

PR #45, opened 2022-07-22
Fix circleci

antomor:
## What

- Add a network to be used in circleci

## Why

- Because it's required to run the integration tests


david-iov (approved):
LGTM

PR #48, opened 2022-08-05
PP-130-v2/remainder management

AndresQuijano:
## What

- Modify the Collector contract to properly manage the remainder. It was done introducing an additional wallet where the remainder can be sent.
- Additionally, in this PR Waffle is introduced as testing framework and the Collector is fully covered with unit testing.

## Why

- Currently one can only update partners when token balance = 0, but due to how integer division works in solidity, it is very likely that there will be a remainder after distributing all fees to the partners. A token amount of 999 that needs to be distributed to 4 partners would leave a remainder of balance = 3 after withdrawal.

## Refs

- [PP-130](https://rsklabs.atlassian.net/browse/PP-130)

### DIFF
--- a/contracts/Collector.sol
+++ b/contracts/Collector.sol
@@ -9,18 +9,21 @@ import "@openzeppelin/contracts/math/SafeMath.sol";
 contract Collector is ICollector{
     IERC20 public token;
     address public owner;
+    address private remainderAddress;
     RevenuePartner[] private partners;
 
     constructor(
         address _owner,
         IERC20 _token,
-        RevenuePartner[] memory _partners
+        RevenuePartner[] memory _partners,
+        address _remainderAddress
     )
     public
     validShares(_partners)
     {   
         owner = _owner;
         token = _token;
+        remainderAddress = _remainderAddress;
         for (uint i = 0; i < _partners.length; i++)
             partners.push(_partners[i]);
     }
@@ -29,18 +32,33 @@ contract Collector is ICollector{
     external
     validShares(_partners)
     onlyOwner()
-    {
-        uint balance = token.balanceOf(address(this));
-        require(balance == 0, "can't update with balance > 0");
-    
+    noBalanceToShare()
+    {    
         delete partners;
-        
+
         for (uint i = 0; i < _partners.length; i++)
             partners.push(_partners[i]);
     }
 
-    function getBalance()
+    //@notice Withdraw the actual remainder and then update the remainder's address
+    //for a new one. This function is the only way to withdraw the remainder.
+    function updateRemainderAddress(address _remainderAddress) 
     external
+    onlyOwner()
+    noBalanceToShare()
+    {
+        address oldRemainderAddress = remainderAddress;
+        remainderAddress = _remainderAddress;
+
+        uint balance = token.balanceOf(address(this));
+
+        if(balance != 0) {
+            token.transfer(oldRemainderAddress, balance);
+        }
+    }
+
+    function getBalance()
+    external view
     returns (uint)
     {
         return token.balanceOf(address(this));
@@ -52,7 +70,7 @@ contract Collector is ICollector{
     onlyOwner()
     {
         uint balance = token.balanceOf(address(this));
-        require(balance > 0, "no revenue to share");
+        require(balance > partners.length, "no revenue to share");
 
         for(uint i = 0; i < partners.length; i++)
             token.transfer(partners[i].beneficiary, SafeMath.div(SafeMath.mul(balance, partners[i].share), 100));
@@ -68,17 +86,27 @@ contract Collector is ICollector{
     }
 
     modifier validShares(RevenuePartner[] memory _partners){
-        uint totalShares = 0;            
-        for(uint i = 0; i < _partners.length; i++)
+        uint totalShares = 0;
+
+        for(uint i = 0; i < _partners.length; i++){
             totalShares = totalShares + _partners[i].share;
+            require(_partners[i].share > 0, "0 is not a valid share");
+        }
 
         require(totalShares == 100, "total shares must add up to 100%");
-
         _;
     }
 
     modifier onlyOwner(){
-        require(msg.sender == owner, "can only call from owner");
+        require(msg.sender == owner, "only owner can call this");
+        _;
+    }
+
+    modifier noBalanceToShare(){
+        uint balance = token.balanceOf(address(this));
+
+        require(balance < partners.length, "there is balance to share");
+
         _;
     }
 }
--- a/contracts/Collector.sol
+++ b/contracts/Collector.sol
@@ -7,10 +7,35 @@ import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/math/SafeMath.sol";
 
 contract Collector is ICollector{
-    IERC20 public token;
-    address public owner;
     address private remainderAddress;
     RevenuePartner[] private partners;
+    IERC20 public token;
+    address public owner;
+
+    modifier validShares(RevenuePartner[] memory _partners){
+        uint256 totalShares;
+
+        for(uint256 i = 0; i < _partners.length; i++){
+            totalShares = totalShares + _partners[i].share;
+            require(_partners[i].share > 0, "0 is not a valid share");
+        }
+
+        require(totalShares == 100, "Total shares must add up to 100%");
+        _;
+    }
+
+    modifier onlyOwner(){
+        require(msg.sender == owner, "Only owner can call this");
+        _;
+    }
+
+    modifier noBalanceToShare(){
+        uint256 balance = token.balanceOf(address(this));
+
+        require(balance < partners.length, "There is balance to share");
+
+        _;
+    }
 
     constructor(
         address _owner,
@@ -24,19 +49,19 @@ contract Collector is ICollector{
         owner = _owner;
         token = _token;
         remainderAddress = _remainderAddress;
-        for (uint i = 0; i < _partners.length; i++)
+        for (uint256 i = 0; i < _partners.length; i++)
             partners.push(_partners[i]);
     }
 
     function updateShares(RevenuePartner[] memory _partners) 
     external
     validShares(_partners)
-    onlyOwner()
-    noBalanceToShare()
+    onlyOwner
+    noBalanceToShare
     {    
         delete partners;
 
-        for (uint i = 0; i < _partners.length; i++)
+        for (uint256 i = 0; i < _partners.length; i++)
             partners.push(_partners[i]);
     }
 
@@ -44,13 +69,13 @@ contract Collector is ICollector{
     //for a new one. This function is the only way to withdraw the remainder.
     function updateRemainderAddress(address _remainderAddress) 
     external
-    onlyOwner()
-    noBalanceToShare()
+    onlyOwner
+    noBalanceToShare
     {
         address oldRemainderAddress = remainderAddress;
         remainderAddress = _remainderAddress;
 
-        uint balance = token.balanceOf(address(this));
+        uint256 balance = token.balanceOf(address(this));
 
         if(balance != 0) {
             token.transfer(oldRemainderAddress, balance);
@@ -59,7 +84,7 @@ contract Collector is ICollector{
 
     function getBalance()
     external view
-    returns (uint)
+    returns (uint256)
     {
         return token.balanceOf(address(this));
     }
@@ -67,46 +92,21 @@ contract Collector is ICollector{
     function withdraw() 
     external 
     override
-    onlyOwner()
+    onlyOwner
     {
-        uint balance = token.balanceOf(address(this));
-        require(balance > partners.length, "no revenue to share");
+        uint256 balance = token.balanceOf(address(this));
+        require(balance > partners.length, "Not enough balance to split");
 
-        for(uint i = 0; i < partners.length; i++)
+        for(uint256 i = 0; i < partners.length; i++)
             token.transfer(partners[i].beneficiary, SafeMath.div(SafeMath.mul(balance, partners[i].share), 100));
     }
 
     function transferOwnership(address _owner)
     external 
     override
-    onlyOwner()
+    onlyOwner
     {
-        require(_owner != address(0), "new owner is the zero address");
+        require(_owner != address(0), "New owner is the zero address");
         owner = _owner;
-    }
-
-    modifier validShares(RevenuePartner[] memory _partners){
-        uint totalShares = 0;
-
-        for(uint i = 0; i < _partners.length; i++){
-            totalShares = totalShares + _partners[i].share;
-            require(_partners[i].share > 0, "0 is not a valid share");
-        }
-
-        require(totalShares == 100, "total shares must add up to 100%");
-        _;
-    }
-
-    modifier onlyOwner(){
-        require(msg.sender == owner, "only owner can call this");
-        _;
-    }
-
-    modifier noBalanceToShare(){
-        uint balance = token.balanceOf(address(this));
-
-        require(balance < partners.length, "there is balance to share");
-
-        _;
     }
 }
--- a/test/collector.test.ts
+++ b/test/collector.test.ts
@@ -49,7 +49,7 @@ async function prepareAll(wallets: Wallet[]) {
     ] = wallets;
 
     const testToken = await deployContract(owner, TestToken);
-    const partners = buildPartners([partner1, partner2, partner3, partner4]);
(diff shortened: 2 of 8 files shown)


ironFe93 (approved):
lgtm

franciscotobar (approved):

antomor (changes requested):
@AndresQuijano @ironFe93 LGTM! The only thing that I noticed is the repetition of the call `await loadFixture(prepareAllFixture);` across all the tests. If possible I'd define a `beforeEach` method in the `describe` sections and I'd move that call there to avoid code repetition.

contracts/Collector.sol
```diff
@@ -29,18 +32,33 @@ contract Collector is ICollector{
     external
     validShares(_partners)
     onlyOwner()
-    {
-        uint balance = token.balanceOf(address(this));
-        require(balance == 0, "can't update with balance > 0");
-    
+    noBalanceToShare()
```
jurajpiar, line 35: you don't the brackets if the modifier doesn't take any arguments

contracts/Collector.sol
```diff
@@ -68,17 +86,27 @@ contract Collector is ICollector{
...
-        require(msg.sender == owner, "can only call from owner");
+        require(msg.sender == owner, "only owner can call this");
+        _;
+    }
+
+    modifier noBalanceToShare(){
+        uint balance = token.balanceOf(address(this));
+
+        require(balance < partners.length, "there is balance to share");
+
```
jurajpiar, line 109: I'd like to use this to propose a standard for an order of definitions, as follows:
1. private storage
2. internal storage
3. public storage
4. modifiers
5. constructor
6. private methods
7. internal methods
8. external methods
9. public methods

contracts/Collector.sol
```diff
@@ -52,7 +70,7 @@ contract Collector is ICollector{
     onlyOwner()
     {
         uint balance = token.balanceOf(address(this));
-        require(balance > 0, "no revenue to share");
+        require(balance > partners.length, "no revenue to share");
```
jurajpiar, line 73:
```suggestion
        require(balance > partners.length, "Not enough left-over balance to split between partners");
```

contracts/Collector.sol
```diff
@@ -68,17 +86,27 @@ contract Collector is ICollector{
     }
 
     modifier validShares(RevenuePartner[] memory _partners){
-        uint totalShares = 0;            
-        for(uint i = 0; i < _partners.length; i++)
+        uint totalShares = 0;
```
jurajpiar, line 89: zero is default state, no need to assign it (reassignment to same value costs gas.).

contracts/Collector.sol
```diff
@@ -68,17 +86,27 @@ contract Collector is ICollector{
...
+
+        for(uint i = 0; i < _partners.length; i++){
             totalShares = totalShares + _partners[i].share;
+            require(_partners[i].share > 0, "0 is not a valid share");
+        }
```
jurajpiar, line 94: for loops are super expensive.. could we use different strategy?

AndresQuijano:
> @AndresQuijano @ironFe93 LGTM! The only thing that I noticed is the repetition of the call `await loadFixture(prepareAllFixture);` across all the tests. If possible I'd define a `beforeEach` method in the `describe` sections and I'd move that call there to avoid code repetition.

I think this my approach is cleaner since I'm taking different elements from that call every time. The code repetition could be avoid defining global variables for the collector, testToken, utilWallet, etc. and then creating a beforeEach() with just one line inside but I think the code will be less clean. [The official documentation](https://ethereum-waffle.readthedocs.io/en/latest/fixtures.html) uses a similar approach as mine. I'm open to discuss it or perhaps have a third opinion. What do you think about this @jurajpiar ?

test/collector.test.ts
```diff
@@ -0,0 +1,351 @@
...
+            share: shares?.[1] ?? SECOND_PARTNER_SHARE
+        },
+        {
+            beneficiary: partners[2].getAddress(),
+            share: shares?.[2] ?? THIRD_PARTNER_SHARE
+        },
+        {
+            beneficiary: partners[3].getAddress(),
+            share: shares?.[3] ?? FOURTH_PARTNER_SHARE
+        }
```
jurajpiar, line 36: surely this could be iterated...

test/collector.test.ts
```diff
@@ -0,0 +1,351 @@
...
+        remainder.address
+    ]);
+
+    return { collector, testToken, partners, remainder, utilWallet };
+}
+
+describe('Collector', () => {
+    async function prepareAllFixture(wallets: Wallet[]) {
+        return prepareAll(wallets);
+    }
```
jurajpiar, line 67: why does this have to be repeated here?

test/collector.test.ts
```diff
@@ -0,0 +1,351 @@
...
+
+    const collector = await deployContract(owner, Collector, [
+        owner.address,
+        testToken.address,
+        partners,
+        remainder.address
+    ]);
+
+    return { collector, testToken, partners, remainder, utilWallet };
+}
```
jurajpiar, line 62: I think I'd rather these were separate functions each doing a single job only.

test/collector.test.ts
```diff
@@ -0,0 +1,351 @@
...
+
+            await testToken.mint(100, collector.address);
+            await collector.withdraw();
+
+            expect(await testToken.balanceOf(partners[0].beneficiary)).to.equal(
+                FIRST_PARTNER_SHARE
+            );
+            expect(await testToken.balanceOf(partners[1].beneficiary)).to.equal(
+                SECOND_PARTNER_SHARE
+            );
```
jurajpiar, line 295: Could you please add expect description wherever you use more than one expect call (at least those of the same types) so that it is clear which one failed if any.

test/collector.test.ts
```diff
@@ -0,0 +1,351 @@
...
+            expect(await testToken.balanceOf(collector.address)).to.equal(0);
+        });
+
+        it('Should fail when no revenue to share', async () => {
+            const { collector, testToken } = await loadFixture(
+                prepareAllFixture
+            );
+            await testToken.mint(1, collector.address);
+            await expect(collector.withdraw()).to.be.revertedWith(
```
jurajpiar, line 310: We shell add this as a formatting rule, but until then, always add a space before await and return statements for clarity

contracts/Collector.sol
```diff
@@ -52,7 +70,7 @@ contract Collector is ICollector{
     onlyOwner()
     {
         uint balance = token.balanceOf(address(this));
-        require(balance > 0, "no revenue to share");
+        require(balance > partners.length, "no revenue to share");
 
         for(uint i = 0; i < partners.length; i++)
             token.transfer(partners[i].beneficiary, SafeMath.div(SafeMath.mul(balance, partners[i].share), 100));
```
jurajpiar, line 76: we should not do this. Each partner should trigger their own withdrawal. the contract should not send value.
See https://docs.soliditylang.org/en/v0.8.15/common-patterns.html#withdrawal-from-contracts

contracts/Collector.sol
```diff
@@ -9,18 +9,21 @@ import "@openzeppelin/contracts/math/SafeMath.sol";
...
+        address _remainderAddress
     )
     public
     validShares(_partners)
     {   
         owner = _owner;
         token = _token;
+        remainderAddress = _remainderAddress;
         for (uint i = 0; i < _partners.length; i++)
             partners.push(_partners[i]);
```
jurajpiar, line 28: so here the constructor iterates over the array again after doing the same job when validating shares.

jurajpiar (changes requested):
Main point of worry is the amount of iterations and the fact that we send value to the partners, instead of them withdrawing.
Also, I'd recommend using  `uint256` instead of `uint`. They mean the same thing (this time this is true), but it makes the dev aware of the size at all time.

test/collector.test.ts
AndresQuijano: I tried to add a custom message for this expect and then make it fail but the massage never appeared. It looks like the should method has this functionality but expect no. So one option is to migrate everything from expect to should. Honestly I think it doesn't worth the effort.

contracts/Collector.sol
AndresQuijano: I think that we should create a specific task for this.

contracts/Collector.sol
AndresQuijano: I think it's out of the scope of this task and we should create a new task for this.

test/collector.test.ts
AndresQuijano: My intention was to keep the logic related to preparation outside of the describe. The fixture should be part of the describe so this is the way to connect both.

test/collector.test.ts
AndresQuijano: Not sure about this. This function groups three single calls to other functions, two of them (deployContract) external functions, and avoid to perform this calls every time in multiple points of the code.

test/collector.test.ts
AndresQuijano: I implemented the iterated version but it was less readable since it implies the use of an array of cons and an extra variable for the size of the array. Those variables where used on a test bellow that was also less readable. I rolled it back to this version.

contracts/Collector.sol
AndresQuijano: Done.

contracts/Collector.sol
AndresQuijano: Done. Let's talk with teammates to establish this as a standard.

contracts/Collector.sol
AndresQuijano: Done with the message we agree on a conversation through Slack.

contracts/Collector.sol
AndresQuijano: Done.

contracts/Collector.sol
AndresQuijano: I think it's out of the scope of this task and we should create a new task for this.

test/collector.test.ts
AndresQuijano: Done.

contracts/Collector.sol
jurajpiar: done: https://www.notion.so/iovlabs/order-of-definition-8173bded80204f55a9bf88c3b3779a65

test/collector.test.ts
antomor: Maybe I'm missing something, but checking the [documentation](https://www.chaijs.com/api/bdd/#method_equal) I think the right way to achieve this is
```suggestion
            expect(await testToken.balanceOf(partners[1].beneficiary)).to.equal(
                SECOND_PARTNER_SHARE,
                'Wrong Partner[1] balance'
            );
```

contracts/Collector.sol
antomor: Created a [task](https://rsklabs.atlassian.net/browse/PP-351) for this.

contracts/Collector.sol
antomor: Created a [task](https://rsklabs.atlassian.net/browse/PP-351) for this.

antomor (commented):
Apart from the currently open conversations, LGTM

test/collector.test.ts
jurajpiar: would it? consider following:
```
const PARNER_SHARES = [20, 30, 40, 10];

type Partner = {
    beneficiary: string,
    share: number,
}

const buildPartners = (wallets: Wallet[], shares: number[]): Partner[] => {
    return wallets.map(({getAddress}: Wallet, partnerIndex) => ({
        beneficiary: getAddress(),
        share: shares?.[partnerIndex] ?? FIRST_PARTNER_SHARE
    }));
```

on the other hand, if you are manually defining these you might as well manually define the whole object without the need for this function in the first place ;)

test/collector.test.ts
jurajpiar: this is absolutely unnecessary. you can use the function directly, even though it is outside of the describe body.

test/collector.test.ts
AndresQuijano: Yes, and that's what I did. But it didn't work.

franciscotobar: I tried the following way and works            
           `expect(
                await testToken.balanceOf(partners[1].beneficiary),
                'Wrong Partner[1] balance'
            ).to.equal(SECOND_PARTNER_SHARE);`

test/collector.test.ts
AndresQuijano: This is a very complex solution for that simple problem. It's more efficient, advanced and elegant but think in simplicity as what can be more understandable for a junior developer. 
The code you propose here is not working for me by the way. I did something in between.

test/collector.test.ts
AndresQuijano: You're right on this. Thank you. Done.

test/collector.test.ts
AndresQuijano: Yes, it was working since the beginning. I didn't noticed that the personalized error was concatenated to the test framework message. My bad.

test/collector.test.ts
jurajpiar: then, as I said above, just hardcode the entire object, no need for functions. Btw:
> It's more efficient, advanced, elegant and shows a lot of knowledge in Typescript but think of simplicity as what can be more understandable for a junior developer
is not a good argument.
1. if something has those properties you layed out, it IS more understandable to junior dev
2. it does not show a lot of knowledge in TS at all. the only ts thing here is one type
3. my suggestion IS more simple
4. even junior developer would be expected to write clean code
5. tests need to follow clean code principals, as they are the single source of truth about the code without looking at the code

test/collector.test.ts
```diff
@@ -11,33 +11,28 @@ import { Wallet } from '@ethersproject/wallet';
...
-            beneficiary: partners[2].getAddress(),
-            share: shares?.[2] ?? THIRD_PARTNER_SHARE
-        },
-        {
-            beneficiary: partners[3].getAddress(),
-            share: shares?.[3] ?? FOURTH_PARTNER_SHARE
-        }
-    ];
+const PARTNER_SHARES = [20, 30, 40, 10];
+const NUMBER_OF_PARTNERS = 4;
```
antomor, line 15: Is there any reason no to do the following?
```suggestion
const NUMBER_OF_PARTNERS = PARTNER_SHARES.length;
```

test/collector.test.ts
```diff
@@ -232,15 +223,25 @@ describe('Collector', () => {
             expect('updateRemainderAddress').to.be.calledOnContract(collector);
         });
 
-        it('Should update remainder when token balance is a remainder and should withdraw remainder', async () => {
+        it.only('Should update remainder when token balance is a remainder and should withdraw remainder', async () => {
```
antomor, line 226: I think we should remove `.only` from here.
```suggestion
        it('Should update remainder when token balance is a remainder and should withdraw remainder', async () => {
```

antomor (changes requested):
Apart from the comments I've just left, I think there are still some open conversations that need to be resolved.

test/collector.test.ts
```diff
@@ -11,33 +11,28 @@ import { Wallet } from '@ethersproject/wallet';
...
-async function prepareAll(wallets: Wallet[]) {
+async function buildPartners(parternWallets: Wallet[], shares?: number[]) {
+    const partners: Partner[] = [];
+
+    for (let i = 0; i < NUMBER_OF_PARTNERS; i++) {
+        partners.push({
+            beneficiary: await parternWallets[i].getAddress(),
+            share: shares?.[i] ?? PARTNER_SHARES[i]
+        });
+    }
```
jurajpiar, line 30: please please prefer map to for loops. Main reasons are:
1. immutability
2. map, filter, etc are optimised, for loop is not

The first reason is obviously the killer of for loops unless you really really REALLY need the performance increase (for small arrays it's pretty much negligible)

test/collector.test.ts
AndresQuijano: Done. Regarding this we should define a process for the cases where reviewer and owner disagree. For this case I just want to close this PR so I implemented an even more complex version of the proposed solution but I don't think it's the best one.

test/collector.test.ts
AndresQuijano: Yes, that's better. Done.

test/collector.test.ts
AndresQuijano: Fixed, my bad.

test/collector.test.ts
AndresQuijano: Done.

antomor (approved):
LGTM

jurajpiar (approved):
LG

PR #54, opened 2022-09-02
PP-325/REL-002

franciscotobar:
## What

- The problem is related to the fact that the initialize method of the template contract is never called, so a malicious attacker could call it and therefore set a logic address. After doing that, the attacker could call any function and that would lead the logic address to be executed. Potentially the logic address could contain a selfdestruct instruction that would destroy the template contract. If so, the proxies contracts won’t be able to execute any transaction anymore.

- I decided to just initialise the CustomSmartWallet implementation contract in the constructor.

## Why

- By using this approach the CustomSmartWallet initialization is blocked since the deployment to prevent malicious initialisation.

## Refs

- [Jira issue](https://rsklabs.atlassian.net/browse/PP-325)

### DIFF
--- a/contracts/smartwallet/CustomSmartWallet.sol
+++ b/contracts/smartwallet/CustomSmartWallet.sol
@@ -14,7 +14,11 @@ contract CustomSmartWallet is IForwarder {
 
     uint256 public override nonce;
     bytes32 public constant DATA_VERSION_HASH = keccak256("2");
-    bytes32 public domainSeparator;    
+    bytes32 public domainSeparator;
+
+    constructor() public {
+        setOwner(msg.sender);
+    }
     
     function buildDomainSeparator() internal {
         domainSeparator = keccak256(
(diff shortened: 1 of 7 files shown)


contracts/smartwallet/CustomSmartWallet.sol
```diff
@@ -14,7 +14,11 @@ contract CustomSmartWallet is IForwarder {
 
     uint256 public override nonce;
     bytes32 public constant DATA_VERSION_HASH = keccak256("2");
-    bytes32 public domainSeparator;    
+    bytes32 public domainSeparator;
+
+    constructor() public {
+        setOwner(msg.sender);
```
ironFe93, line 20: Wouldn't caling setOwner() here make it impossible to run initialize() afretwards?

ironFe93 (commented):
left a comment, if responded I can approve.

contracts/smartwallet/CustomSmartWallet.sol
franciscotobar: By initializing it, we are initializing the logic contract and not the proxy which in our case are the smart wallet address created with the index. 

Im still doing to testing to make sure that its working properly but so far so good. I just need to confirm if we need to setup to msg.sender or a different address or if we add an initialization flag, but that increase the gas cost.

franciscotobar:
I tried to use expect and assert, specifically the expect().to.be.rejectedWith and assert.isRejected. It seems to be that it cannot be combined.

antomor (approved):
@franciscotobar Thank you for these changes, LGTM!

PR #60, opened 2022-09-16
Fix patch package

antomor:
## What

- Fix the patch applied to @openeth/truffle-typings
- Also include the package-lock.json file because it's required by patch-packages to create new patches.

## Why

- It raises an error when it tries to apply it, due to a manual change in the file.


ironFe93 (approved):
LGTM

PR #68, opened 2022-09-28
PP473: Replace Ownable with own ownership mechanism

AndresQuijano:
## What

- Creates an abstract contract called Ownable and replaces all the implementations of Openzeppelin/Ownable

## Why

- The Openzeppelin implementation has not required functionalities that uses extra gas

## Refs

- [PP-473](https://rsklabs.atlassian.net/browse/PP-473)

### DIFF
--- a/contracts/Ownable.sol
+++ b/contracts/Ownable.sol
@@ -23,9 +23,8 @@ abstract contract Ownable {
      * @dev Initializes the contract setting the deployer as the initial owner.
      */
     constructor () internal {
-        address msgSender = msg.sender;
-        _owner = msgSender;
-        emit OwnershipTransferred(address(0), msgSender);
+        _owner = msg.sender;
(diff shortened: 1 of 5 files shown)


contracts/Ownable.sol
```diff
@@ -0,0 +1,67 @@
...
+abstract contract Ownable {
+    address private _owner;
+
+    event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
+
+    /**
+     * @dev Initializes the contract setting the deployer as the initial owner.
+     */
+    constructor () internal {
+        address msgSender = msg.sender;
```
franciscotobar, line 26: Ill avoid using an extra variable here, to reduce cost usage.

franciscotobar (approved):
LGTM

jurajpiar (approved):
LG

PR #79, opened 2022-10-19
PP-509 remove tokens from acceptedTokens array

ironFe93:
## What

- implement removeToken()

## Why

- A method to delist tokens from rif-relay was needed

PR #80, opened 2022-10-20
PP-509: add the removeToken method

ironFe93:
## What

method to remove tokens from acceptedTokens array

## Why

Functionally there needs to exist a counterpart to acceptToken()

## Refs

tests contained in:
https://github.com/rsksmart/rif-relay-contracts/pull/79


antomor (approved):
LGTM. 
Tests aren't included here because we decided to include them directly in the new version, see #79

PR #83, opened 2022-10-23
PP 429: Rewrite custom smart wallet tests

antomor:
## What

- Rewrite the unit tests for CustomSmartWallet contract

## Why

- Modernization of the project by switching to Ethers.js and Hardhat

## Refs

- [PP-429](https://rsklabs.atlassian.net/browse/PP-429)


test/smartwallet/customSmartWallet.test.ts
```diff
@@ -733,3 +733,564 @@
 // );
 
 // /* TODO - ADD tests for recover method */
```
antomor, line 735: @AndresQuijano I left the whole file as it was before to let you check if additional tests have been added. Could you please review it and eventually add the missing cases and afterward delete the commented part?

PR #71, opened 2022-10-06
feature(upgrade): expose and clean dist

jurajpiar:
## What

- adds separate tsconfig for building
- organises scripts in package.json
- includes only dist folder in the package (adds files property to package.json)
- removes tests from dist folder
- adds index file to root

## Why

- this should improve usability of this package
- removes the need to navigate to dist folder in imports

## Refs

- no ticket 😬 sorry

PR #87, opened 2022-11-03
Expose dist

antomor:
## What

- Expose the dist folders to be used by other libraries

## Why

- Otherwise the other libraries (common, client, server) cannot make use of it

## Refs

- It replaces PR #71


tsconfig.build.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": [
+    "test",
+    "tasks"
+  ]
+}
+
```
jurajpiar, line 8: add trailing line or remove space from the last line

.eslintrc.cjs
```diff
@@ -51,6 +51,19 @@ module.exports = {
...
   },
+  // FIXME: we need to fix those rules in the test files
+  overrides: [
+    {
+      files: ["test/**"],
+      rules: {
+        "@typescript-eslint/no-unsafe-assignment": "warn",
+        "@typescript-eslint/no-unsafe-call": "warn",
+        "@typescript-eslint/no-unsafe-argument": "warn",
+        "@typescript-eslint/no-unsafe-return": "warn",
```
jurajpiar, line 64: why are these not errors? is that to make the tests pass as they are?

jurajpiar (approved):
just check out your trailing lines

franciscotobar:
The dist folder it is not being included in the build.

.eslintrc.cjs
antomor: Yes, exactly, I'm going to create a specific task to work on this (fix eslint and tsc errors on tests)

antomor:
> The dist folder it is not being included in the build. I think that is not being generated in the github flow.

@franciscotobar thank you for highlighting this. I've checked, and it looks like the issue is related to [those lines](https://github.com/rsksmart/rif-relay-contracts/blob/feature/ethers/exposeDist/scripts/prepare.npm#L7-L8). Right now, for some reasons, we are deleting the generated dist after creating it (at installation time, in the prepare script). Furthermore, I've noticed that the command `rsync -avzh --progress dist .` doesn't have any effect at all (if we wanted to copy the folder content we should have run `rsync -avzh --progress dist/ .`), so I'm assuming we could delete those lines. @jurajpiar could you confirm?

antomor:
> just check out your trailing lines

@jurajpiar thank this, I've added an eslint rule to solve it 👍🏽

index.ts
```diff
@@ -0,0 +1,4 @@
+import * as deployment from './scripts/deploy';
+import * as types from './typechain-types';
+
+export { types, deployment };
```
franciscotobar, line 4: sorry this was my fault. I just did some testing and we need to change this whole file with this:

```suggestion
export * from './typechain-types'; 
export * from './scripts/deploy';
```

franciscotobar: well or depends, why do we need to export the deploy scripts? Are we going to use outside contracts?

antomor: @franciscotobar why do we want to change it? what's the issue exactly?

franciscotobar: By keeping it in that way, we are exporting it like this:

`
import {
  deployment,
  types
} from '@rsksmart/rif-relay-contracts';
`

So to access to each typing we need to do:
`types.DeployVerifier__factory`

franciscotobar: Before it was working because we were accessing to the src files and now since the src is not included, we cannot access like we where doing it before, by doing that change can do this:

`import {
  DeployVerifier__factory,
} from '@rsksmart/rif-relay-contracts';`

antomor: @franciscotobar good point, thank you for pointing it out 🙏🏽

antomor:
@franciscotobar I've also added the changes that you applied to `feature/ethers/exposeDist-v2`. If you confirm that this PR is ok, i'm going to merge it and delete the branch `feature/ethers/exposeDist-v2`

PR #89, opened 2022-11-08
PP-533: refactor verifiers

AndresQuijano:
## What

- Refactor verifiers so all the common logic is in an abstract class

## Why

- Simplify code and avoid code repetition

## Refs

- [PP-533](https://rsklabs.atlassian.net/browse/PP-533)


antomor (changes requested):
@AndresQuijano LGTM.
The only doubt is: why did we delete the `ITokenHandler`?

PR #95, opened 2022-11-16
PP-566: migrate script deploy collector

AndresQuijano:
## What

- Migrates the script to deploy to collector to use Hardhat and Ethers.js

## Why

- Modernization of the project

## Refs

-[PP-566](https://rsklabs.atlassian.net/browse/PP-566)


antomor (changes requested):

hardhat.config.ts
```diff
@@ -103,4 +104,15 @@ task('remove-tokens', 'Removes a list of tokens')
   }
 );
 
+task('deploy-collector', 'Deploys the collector')
+  .addOptionalPositionalParam('collectorConfigFileName', 'Relative path to an alternative config file')
+  .addOptionalPositionalParam('outputFileName', 'Relative path to the output file')
+  .setAction(async (taskArgs: { 
+    collectorConfigFileName: string,
+    outputFileName: string
+  }, hre) => {
```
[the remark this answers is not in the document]
antomor: fixed

package.json
```diff
@@ -67,6 +67,7 @@
     "postclean": "hardhat clean",
     "compile": "npm run clean && hardhat compile --force",
     "deploy": "hardhat deploy --network",
+    "deploy-collector": "hardhat deploy-collector --network",
```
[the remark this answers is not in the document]
antomor: fixed

hardhat.config.ts
```diff
@@ -103,4 +104,15 @@ task('remove-tokens', 'Removes a list of tokens')
   }
 );
 
+task('deploy-collector', 'Deploys the collector')
+  .addOptionalPositionalParam('collectorConfigFileName', 'Relative path to an alternative config file')
+  .addOptionalPositionalParam('outputFileName', 'Relative path to the output file')
```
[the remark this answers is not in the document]
antomor: fixed

antomor:
All the changes to the solidity contracts are related to formatting rules. No logic has been changed there.

scripts/deployCollector.ts
```diff
@@ -0,0 +1,159 @@
...
+  const { collectorOwner, partners, tokenAddress, remainderAddress } =
+    inputConfig;
+
+  const collectorFactory = await ethers.getContractFactory('Collector');
+  const collector = await collectorFactory.deploy(
+    collectorOwner,
+    tokenAddress,
+    partners,
+    remainderAddress
+  );
```
jurajpiar, line 66: Not important, just an FYI: you should be able to use `Collector_factory` class from type-chain directly instead of specifying it by name string.

```suggestion
  const collectorFactory = new Collector__factory();
  const collector = await collectorFactory.deploy(
    collectorOwner,
    tokenAddress,
    partners,
    remainderAddress
  );
```

package.json
```diff
@@ -65,6 +65,7 @@
     "ci:test": "npm run test",
     "clean": "rm -rf typechain-types cache bin dist artifacts contracts-exposed",
     "postclean": "hardhat clean",
+    "collector:deploy": "hardhat deploy-collector",
```
jurajpiar, line 68: probably better the other way.
```suggestion
    "deploy:collector": "hardhat deploy-collector",
```

scripts/deployCollector.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  if (!network) {
+    throw new Error('Unknown Network');
+  }
+
+  let collectorConfigFileName = taskArgs.collectorConfigFileName;
+
+  if (!collectorConfigFileName) {
+    console.warn(
+      "Missing '--collectorConfig' parameter, 'deploy-collector.input.json' will be used"
```
AndresQuijano, line 44: The script is asking for --collectorConfig parameter but the real name of the parameter is --collector-config-file-name

test/scripts/deployCollector.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        const partners = await collector.getPartners();
+        partners.forEach(([beneficiary, share]) => {
+          expect(collectorConfiguration.partners).to.deep.include({
+            beneficiary,
+            share,
+          });
+        });
+      });
+
+      describe('', function () {
```
AndresQuijano, line 137: Is this describe description missing?

package.json
antomor: Not sure about it. I was thinking about grouping all the collector related scripts, so we would have something like: `collector:deploy`, `collector:withdraw`, `collector:change-partners`. WDYT?

scripts/deployCollector.ts
```diff
@@ -0,0 +1,159 @@
...
+  );
+  console.log(
+    `| Collector Token                             | ${await collector.token()}       |`
+  );
+  console.log(
+    `| Collector Remainder                         | ${remainderAddress}       |`
+  );
+  console.log(
+    '|=============================================|==================================================|\n'
+  );
```
jurajpiar, line 104: why not use `console.table`?

hardhat.config.ts
```diff
@@ -103,4 +104,12 @@ task('remove-tokens', 'Removes a list of tokens')
   }
 );
 
+task('deploy-collector', 'Deploys the collector')
+  .addOptionalParam('collectorConfigFileName', 'Path to an alternative config file')
+  .addOptionalParam('outputFileName', 'Path to the output file')
+  .setAction(async (taskArgs: DeployCollectorArg, hre) => {
+    await deployCollector(taskArgs, hre);
+  }
+);
```
jurajpiar, line 113: I would consider generalising this into a single deploy task. this way we could utilise the hardhat cli powers.

scripts/deployCollector.ts
antomor: you're right, thank you for highlighting this 😅

scripts/deployCollector.ts
antomor: I think that the usage of `await ethers.getContractFactory('Collector');` is easier than `new Collector__factory(contractInterface: ContractInterface, bytecode: BytesLike | { object: string }, signer?: Signer)` and I don't see any benefit in using it. Am i missing anything?

scripts/deployCollector.ts
antomor: Because of the `partners` field. Since its value is an array, the result is very ugly, because it considers each value of the array as a column, so we end up having a table with 4 very large columns, not so readable IMHO.

hardhat.config.ts
ironFe93: I have thought about this as well, but I think it falls out of scope for what this task requires.

scripts/deployCollector.ts
ironFe93: Agreed on the usage of console.table(), please have a look at this: https://github.com/rsksmart/rif-relay-contracts/blob/feature/PP-278/PP-329/upgrade_contracts/scripts/deploy.ts#L141

Readme.md
```diff
@@ -64,13 +64,18 @@ Parameters:
 
 #### Collector Deployment
 
-To deploy a collector, we need to run the script `deploy-collector.js`; it receives two parameters:
-- `collectorConfig`: used to specify the collector owner and the partners configuration (addresses and shares). If not specified, the file `deploy-collector.input.json` will be used. Please have a look at `deploy-collector.input.sample.json` for a sample file.
-- `outputFile`: used to log the main information of the collector deployment. If not specified, the file `revenue-sharing-addresses.json` will be used. 
+To deploy a collector, we need to the script `collector:deploy`. It receives the following parameters:
+- `collector-config-file-name`: optional, used to specify the collector owner and the partners configuration (addresses and shares). If not specified, the file `deploy-collector.input.json` will be used. Please have a look at `deploy-collector.input.sample.json` for a sample file.
```
franciscotobar, line 68: Its not simple to call it, config-file-name?

```suggestion
- `config-file-name`: optional, used to specify the collector owner and the partners configuration (addresses and shares). If not specified, the file `deploy-collector.input.json` will be used. Please have a look at `deploy-collector.input.sample.json` for a sample file.
```

scripts/deployCollector.ts
```diff
@@ -0,0 +1,159 @@
...
+
+interface CollectorConfig {
+  collectorOwner: string;
+  partners: Partner[];
+  tokenAddress: string;
+  remainderAddress: string;
+}
+
+export type DeployCollectorArg = {
+  collectorConfigFileName?: string;
```
franciscotobar, line 22: Its not simple to call it, configFileName?
```suggestion
  configFileName?: string;
```

scripts/deployCollector.ts
```diff
@@ -0,0 +1,159 @@
...
+) => {
+  const { ethers, network } = hre;
+
+  if (!network) {
+    throw new Error('Unknown Network');
+  }
+
+  let collectorConfigFileName = taskArgs.collectorConfigFileName;
+
+  if (!collectorConfigFileName) {
```
franciscotobar, line 42: Its not simple to call it, configFileName?
```suggestion
  if (!configFileName) {
```

scripts/deployCollector.ts
```diff
@@ -0,0 +1,159 @@
...
+  taskArgs: DeployCollectorArg,
+  hre: HardhatRuntimeEnvironment
+) => {
+  const { ethers, network } = hre;
+
+  if (!network) {
+    throw new Error('Unknown Network');
+  }
+
+  let collectorConfigFileName = taskArgs.collectorConfigFileName;
```
franciscotobar, line 40: Its not simple to call it, configFileName?
```suggestion
  let configFileName = taskArgs.configFileName;
```

ironFe93 (changes requested):
--network option is missing in the package.json

package.json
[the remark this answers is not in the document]
antomor: I don't think that's true. AFAIK we can always specify the network parameter with `npx hardhat collector:deploy --network regtest --param1 "<param1>" etc...`
Furthermore, if we don't specify it there, we could also use the default network.

scripts/deployCollector.ts
antomor: Yes, I managed to solve the issue I had with the partners by creating an object specifically to be printed.

package.json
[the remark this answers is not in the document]
ironFe93: ok you have to send in an empty -- first otherwise it will identify as positional parameters.

ironFe93 (approved):

hardhat.config.ts
antomor: I've actually tried to use the subtasks, but it looks like they're particularly useful when we want to split a task into multiple actions. For our specific use case, I think it complicates the code without any benefit.

antomor:
@franciscotobar: I applied the renaming of the input param.
@ironFe93: I decided to remove some tasks from the package.json because there was no point to have them there.
@jurajpiar I changed the way we log the output information, but we still use the old way to build the factory. Some tests were failing and that change may require some additional investigation.
Could you please review the changes? 🙏🏼

package.json
```diff
@@ -65,9 +63,9 @@
...
     "compile": "npm run clean && hardhat compile --force",
     "deploy": "hardhat deploy --network",
```
ironFe93, line 68: Agree with not cluttering the package json with scripts. Shouldn't the 'deploy' script also be removed from here?

ironFe93 (approved):
agree on removing stuff from package json

package.json
antomor: I wasn't unsure about it. WDYT?

ironFe93: It falls under the same category as all other tasks, so yes, we could remove it.

jurajpiar (approved):
LG

antomor (approved):

PR #96, opened 2022-11-17
PP-355/update-rif-relay-server

franciscotobar:
## What

- Expose necessary types for server

## Why

- To use typechain types in the server

## Refs

- [PP-355](https://rsklabs.atlassian.net/browse/PP-355)


antomor (approved):
@franciscotobar I've realized we had some prettier/eslint issues, so I had to ignore the dist folder. Apart from this, LGTM

package.json
```diff
@@ -70,11 +70,17 @@
...
-    "format:ts": "prettier -c -w --no-editorconfig **/*.ts",
+    "format:fix": "npm run format -- --write",
+    "format:fix:sol": "npm run format:sol -- --write",
+    "format:fix:ts": "npm run format:ts -- --write",
+    "format:sol": "prettier --list-different --no-editorconfig \"{,contracts/**/}*.sol\"",
+    "format:ts": "prettier --list-different  --no-editorconfig \"{,test/**/}*.ts\" \"{,scripts/**/}*.ts\" index.ts",
     "lint": "npm run lint:sol && npm run lint:ts",
-    "lint:sol": "solhint --fix 'contracts/**/*.sol'",
-    "lint:ts": "eslint test --ext .ts --fix",
+    "lint:fix": "npm run lint -- --fix",
```
antomor, line 79: I don't think this works as expected because the `--fix` will be used in the part of the command only and the following command will be run:
```
npm run lint:sol && npm run lint:ts --fix
```

package.json
```diff
@@ -70,11 +70,17 @@
     "deploy-configure": "hardhat deploy --network",
     "docs": "hardhat docgen",
     "format": "npm run format:sol && npm run format:ts",
-    "format:sol": "prettier -c -w --no-editorconfig contracts/**/*.sol contracts/**/**/*.sol",
-    "format:ts": "prettier -c -w --no-editorconfig **/*.ts",
+    "format:fix": "npm run format -- --write",
```
antomor, line 73: I don't think this works as expected because the `--write` will be used in the second part of the script only:
```
npm run format:sol && npm run format:ts -- write
```

antomor (changes requested):
@franciscotobar I think a rebase may be required. However, it looks like that both the contracts and test folders have been formatted only and no code logic is changed there, could you confirm?

package.json
franciscotobar: let me check it, I tried but I did not realize if it was applied to both

package.json
franciscotobar: let me check it, I tried but I did not realize if it was applied to both

package.json
franciscotobar: updated

package.json
franciscotobar: updated

antomor (approved):
LGTM

PR #97, opened 2022-11-18
PP-569: migrate change partners

antomor:
## What

- Migrate the change-partners script to ethers.js

## Why

- It still uses web3.js


antomor:
Although the PR should be ready, this PR depends on #95.

hardhat.config.ts
```diff
@@ -103,4 +104,13 @@ task('remove-tokens', 'Removes a list of tokens')
   }
 );
 
+task('change-partner-shares', 'Change partner shares of a collector')
+  .addParam('collectorAddress', 'address of the collector we want to modify')
+  .addParam('partnerConfig', 'path of the file that includes the partner shares configuration')
+  .addOptionalParam('gasLimit', 'gasLimit to be used for the transaction')
```
ironFe93, line 110: very nice

scripts/utils.ts
```diff
@@ -0,0 +1,19 @@
+import fs from 'node:fs';
+
+export const parseJsonFile = <T>(filePath: string, raiseError = true) => {
+  try {
+    if (fs.existsSync(filePath)) {
+      return JSON.parse(fs.readFileSync(filePath, { encoding: 'utf-8' })) as T;
+    }
+    if (raiseError) {
+      throw new Error(`The file ${filePath} doesn't exist`);
+    }
```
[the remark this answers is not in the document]
antomor: Honestly, I don't like it at all, I just kept this "safe" behaviour because I wanted to use the same code for the `getExistingConfig` in the deploy script. I'll review it again to understand if I can get rid of it.

ironFe93 (approved):
LGTM though I would have probably just thrown an error in the parseJsonFile function.

scripts/changePartnerShares.ts
```diff
@@ -0,0 +1,45 @@
...
+    partnerConfig,
+    gasLimit = defaultTxGas,
+  }: ChangePartnerSharesArg,
+  hre: HardhatRuntimeEnvironment
+) => {
+  const { ethers } = hre;
+
+  const parsedPartnerConfig = parseJsonFile<PartnerConfig>(
+    partnerConfig
+  ) as PartnerConfig;
```
franciscotobar, line 30: We are forcing the type here since we are assuming that it will be failing instead of returning an undefined value, right?

scripts/changePartnerShares.ts
```diff
@@ -0,0 +1,45 @@
...
+  const collector = await ethers.getContractAt('Collector', collectorAddress);
+
+  try {
+    await collector.updateShares(parsedPartnerConfig.partners, { gasLimit });
```
ironFe93, line 34: Quick question, does this not need the 'from' parameter specified? I mention this because the contract method restricts that only the contract owner can call it.

jurajpiar: not in ethers.

scripts/changePartnerShares.ts
antomor: We don't need it actually, thank you for pointing it out.

scripts/changePartnerShares.ts
antomor: @ironFe93 That's a good point, I'll create a task to be able to specify an account at the configuration level.

jurajpiar (commented):
LG

PR #103, opened 2022-11-28
feat(ethers): adds custom signers config

jurajpiar:
## What

- TBD

## Why

- When we deploy the contracts or run the scripts, we want to specify which pre-funded addresses will be used as signers (the same way we’re doing it with truffle).

## Refs

- [Ticket](https://rsklabs.atlassian.net/browse/PP-597)
- [Parent](https://github.com/rsksmart/rif-relay-contracts/pull/55)


antomor:
Closed in favour of #106 because the branch name wasn't correct, and also we adopted a simpler approach

jurajpiar:
Fair enough

PR #104, opened 2022-11-29
Expiration Mechanism (Ethers)

ironFe93:
## What

Adds an expiration mechanism to relay/deploy requests

## Why

- To prevent malicious users to exploit changes in token prices.

## Refs

-

### DIFF
--- a/contracts/Collector.sol
+++ b/contracts/Collector.sol
@@ -71,12 +71,15 @@ contract Collector is ICollector {
         uint256 balance = token.balanceOf(address(this));
         address tokenAddr = address(token);
 
-        if(balance != 0) {
+        if (balance != 0) {
             // solhint-disable-next-line avoid-low-level-calls
-            (bool success, bytes memory ret ) = tokenAddr.call{gas: 200000}(abi.encodeWithSelector(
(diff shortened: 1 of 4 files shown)


contracts/smartwallet/CustomSmartWallet.sol
```diff
@@ -162,6 +162,11 @@ contract CustomSmartWallet is IForwarder {
         require(msg.sender == req.relayHub, "Invalid caller");
 
         _verifySig(suffixData, req, sig);
+        // solhint-disable-next-line not-rely-on-time
+        require(
+            req.validUntilTime == 0 || req.validUntilTime > block.timestamp,
```
AndresQuijano, line 167: Why is 0 a valid value here?

AndresQuijano (commented):
Let's discuss the 0 condition in the wallets.

franciscotobar:
Shouldn't we include some tests with scenarios of expired transactions?

ironFe93:
@franciscotobar I did include a test for the expiration case. It got lost in the rebase/cherry-pick process. Also I realize this PR is messed up. Back to draft it is :)

PR #106, opened 2022-12-02
PP-597: Change the configuration to use custom signers to run the script

antomor:
## What

- Change the configuration to read the custom signers to use from the environment, either by using an env variable or a `.env` file

## Why

- Without it, it's not possible to deploy to Testnet/Mainnet, because there wouldn't be any signer with balance

## How to test

1. Either create a `.env` file or set the following env variables with a private key of an account you've control of ([Here](https://metamask.zendesk.com/hc/en-us/articles/360015289632-How-to-export-an-account-s-private-key) a guide on how to export a private key from metamask)
2. Run the command `npx hardhat console --network testnet`
3. Run a simple transaction:
```
account  = (await ethers.getSigners())[0];
tx = await account.sendTransaction({to: 'address_will_receive_trbtc', value: ethers.utils.parseEther('0.01')});
```
4. Get the `tx.hash` value and wait until the transaction is mined.
5. Check the transaction is executed correctly.

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -1,19 +1,34 @@
 import '@nomicfoundation/hardhat-toolbox';
+import '@nomiclabs/hardhat-ethers';
+import dotenv from 'dotenv';
 import 'hardhat-contract-sizer';
 import 'hardhat-docgen';
 import 'hardhat-watcher';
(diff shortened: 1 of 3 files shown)


franciscotobar (approved):
I just tested it and its working properly.

PR #112, opened 2023-01-05
PP-358: include the contracts in dist

antomor:
## What

- Include the contracts folder in dist

## Why

- We need to use them in the tests repo


package.json
```diff
@@ -56,7 +56,7 @@
   "scripts": {
     "prebuild": "npm run clean",
     "build": "hardhat compile --force",
-    "postbuild": "tsc --project tsconfig.build.json",
+    "postbuild": "tsc --project tsconfig.build.json && cp -R contracts dist/",
```
franciscotobar, line 59: shouldn't be better to remove it from the `.npmignore`, so we can include it, but not in the dist folder?

PR #117, opened 2023-01-31
PP-642/implement-new-sw

franciscotobar:
## What

- Create a new SW that doesn’t automatically transfer RBTC to the owner (EOA)

## Why

- Extra implementation needed by the  User Services team 

## Refs

- [PP-642](https://rsklabs.atlassian.net/browse/PP-662)

### DIFF
--- a/contracts/interfaces/IForwarder.sol
+++ b/contracts/interfaces/IForwarder.sol
@@ -50,6 +50,11 @@ interface IForwarder {
         bytes calldata signature
     ) external view;
 
+     /**
+     * returns the encrypted owner that deployed smart wallet.
+     */
+    function getOwner() external view returns (bytes32 owner);
+
     /**
      * execute a transaction
      * @param forwardRequest - all transaction parameters
--- a/contracts/smartwallet/CustomSmartWallet.sol
+++ b/contracts/smartwallet/CustomSmartWallet.sol
@@ -57,7 +57,7 @@ contract CustomSmartWallet is IForwarder {
         }
     }
 
-    function getOwner() private view returns (bytes32 owner){
+    function getOwner() public view override returns (bytes32 owner){
         assembly {
             owner := sload(
                 0xa7b53796fd2d99cb1f5ae019b54f9e024446c3d12b483f733ccc62ed04eb126a
--- a/contracts/smartwallet/NativeHolderSmartWallet.sol
+++ b/contracts/smartwallet/NativeHolderSmartWallet.sol
@@ -9,15 +9,7 @@ import "./SmartWallet.sol";
 
 contract NativeHolderSmartWallet is SmartWallet {
     
-     /* solhint-disable no-unused-vars */
-    function directExecute(
-        address to,
-        bytes calldata data
-    ) external payable override returns (bool success, bytes memory ret) {
-        revert("Method not implemented");
-    }
-
-    function directExecute(
+    function directExecuteWithValue(
         address to,
         uint256 value,
         bytes calldata data
--- a/contracts/smartwallet/NativeHolderSmartWallet.sol
+++ b/contracts/smartwallet/NativeHolderSmartWallet.sol
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier:MIT
+pragma solidity ^0.6.12;
+pragma experimental ABIEncoderV2;
+
+import "./SmartWallet.sol";
+
+/* solhint-disable no-inline-assembly */
+/* solhint-disable avoid-low-level-calls */
+
+contract NativeHolderSmartWallet is SmartWallet {
+    
+     /* solhint-disable no-unused-vars */
+    function directExecute(
+        address to,
+        bytes calldata data
+    ) external payable override returns (bool success, bytes memory ret) {
+        revert("Method not implemented");
+    }
+
+    function directExecute(
+        address to,
+        uint256 value,
+        bytes calldata data
+    ) external payable returns (bool success, bytes memory ret) {
+        //Verify Owner
+        require(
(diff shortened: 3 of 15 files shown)


contracts/interfaces/IForwarder.sol
```diff
@@ -50,6 +50,11 @@ interface IForwarder {
         bytes calldata signature
     ) external view;
 
+     /**
+     * returns the encrypted owner that deployed smart wallet.
+     */
+    function getOwner() external view returns (bytes32 owner);
```
franciscotobar, line 56: This change is already included in the ethers branch, so I just include it, here.

contracts/smartwallet/CustomSmartWallet.sol
```diff
@@ -57,7 +57,7 @@ contract CustomSmartWallet is IForwarder {
         }
     }
 
-    function getOwner() private view returns (bytes32 owner){
+    function getOwner() public view override returns (bytes32 owner){
```
franciscotobar, line 60: This change is already included in the ethers branch, so I just include it, here.

contracts/smartwallet/SmartWallet.sol
```diff
@@ -107,7 +107,7 @@ contract SmartWallet is IForwarder {
         }
     }
 
-    function directExecute(address to, bytes calldata data) external override payable returns (
+    function directExecute(address to, bytes calldata data) external virtual override payable returns (
```
franciscotobar, line 110: Included the virtual, so it can be overriden.

contracts/smartwallet/SmartWallet.sol
```diff
@@ -199,7 +200,7 @@ contract SmartWallet is IForwarder {
         bytes32 suffixData,
         ForwardRequest memory req,
         bytes memory sig
-    ) private view {
+    ) internal view {
```
franciscotobar, line 203: this is my main concern, the private methods are not copied from the father and you cannot and cannot be overriden, we could either create the same method with a different name or keep it with internal, or maybe Im missing something.

contracts/smartwallet/NativeHolderSmartWallet.sol
```diff
@@ -0,0 +1,78 @@
...
+/* solhint-disable avoid-low-level-calls */
+
+contract NativeHolderSmartWallet is SmartWallet {
+    
+     /* solhint-disable no-unused-vars */
+    function directExecute(
+        address to,
+        bytes calldata data
+    ) external payable override returns (bool success, bytes memory ret) {
+        revert("Method not implemented");
```
antomor, line 17: What do you think if we don't override this method here and we keep the same behaviour that the SmartWallet has?

test/smartwallet/nativeHolderSmartWallet.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    signRequest
+} from '../utils';
+
+use(chaiAsPromised);
+
+const NativeSmartWallet = artifacts.require('NativeHolderSmartWallet');
+const TestToken = artifacts.require('TestToken');
+const TestRecipient = artifacts.require('TestRecipient');
+
+contract.only('NativeHolderSmartWallet', ([worker, fundedAccount]) => {
```
antomor, line 28: we should remove `only` from here

antomor (commented):
@franciscotobar thank you for the changes. I have two main doubts:
1. do we need the `getOwner` function to be public now?
2. why do we override `directExecute`? I wouldn't override it unless we want to change its behaviour.

contracts/interfaces/IForwarder.sol
franciscotobar: not public, but we need to change its accessibility so I decided to do the same change that we applied before.

contracts/smartwallet/CustomSmartWallet.sol
franciscotobar: not public, but we need to change its accessibility so I decided to do the same change that we applied before.

contracts/smartwallet/NativeHolderSmartWallet.sol
franciscotobar: The only reason that I did its to avoid confusion, this contract will have two methods with the name of direct execute, I can leave both, but in one will return the native currency and in the other will not.

test/smartwallet/nativeHolderSmartWallet.test.ts
franciscotobar: sorry, my bad I still need to add the missing tests but I wanted to check functionality

contracts/smartwallet/NativeHolderSmartWallet.sol
antomor: As you discovered, due to the typechain version that we are using, we cannot have function overloading here, so we agreed on having 2 different functions we shouldn't have confusion, right?

franciscotobar: I already remove it :)

test/smartwallet/nativeHolderSmartWallet.test.ts
```diff
@@ -0,0 +1,262 @@
...
+
+            expect(expectedRecipientBalance.toString()).to.be.equal(
+                recipientBalanceAfterExecution.toString()
+            );
+            expect(expectedSwBalance.toString()).to.be.equal(
+                swBalanceAfterExecution.toString()
+            );
+        });
+
+        it.only('Should transfer native currency and execute transaction', async () => {
```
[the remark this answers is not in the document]
franciscotobar: updated

antomor (approved):
LGTM

PR #119, opened 2023-02-07
feat(collector): add multi-token support

jurajpiar:
## What

1. ~Creates new multi-token collector contract~ Modifies Collector contract to allow multiple tokens
2. adds generated files to gitignore
3. removes unnecessary loops in the Collector contract
4. enables compilation in the TDD mode
5. adds contracts and tasks to the watch mode

## Refs

- [PP-664](https://rsklabs.atlassian.net/browse/PP-664)


franciscotobar:
while executing the `npm run test`, Im getting the following warning `Failed to generate 2 stack traces. Run Hardhat with --verbose to learn more.`, before this warning was not showing.

contracts/Collector.sol
```diff
@@ -37,17 +25,26 @@ contract Collector is ICollector {
         _;
     }
 
+    modifier updateValidShares(RevenuePartner[] memory partners) {
+        _;
+        uint256 totalShares;
```
franciscotobar, line 30: why do we have this line of code here?

contracts/MultitokenCollector.sol
```diff
@@ -0,0 +1,136 @@
...
+        address remainderAddress
+    ) external onlyOwner noBalanceToShare {
+        for (uint256 i = 0; i < _tokens.length; i++) {
+            IERC20 token = _tokens[i];
+            uint256 balance = token.balanceOf(address(this));
+
+            if (balance != 0) {
+                // solhint-disable-next-line avoid-low-level-calls
+                (bool success, bytes memory ret) = address(token).call{
+                    gas: 200000
```
franciscotobar, line 77: is it necessary to have this hardcoded value here?

contracts/MultitokenCollector.sol
```diff
@@ -0,0 +1,136 @@
...
+
+    function withdrawToken(IERC20 token) public onlyOwner {
+        uint256 balance = token.balanceOf(address(this));
+        require(balance >= _partners.length, "Not enough balance to split");
+
+        address tokenAddr = address(token);
+
+        for (uint256 i = 0; i < _partners.length; i++) {
+            // solhint-disable-next-line avoid-low-level-calls
+            (bool success, bytes memory ret) = tokenAddr.call(
```
franciscotobar, line 111: why are we using a low level call? Its not better to use a direct call?

```suggestion
           bool success = token.transfer(_partners[i].beneficiary, SafeMath.div(SafeMath.mul(balance, _partners[i].share), 100));
```

test/MultitokenCollector.test.ts
```diff
@@ -0,0 +1,409 @@
...
+
+const buildPartners = async (
+  partnerWallets: SignerWithAddress[],
+  shares: number[]
+) => {
+  return Promise.all(
+    partnerWallets.map(async (wallet, index) => {
+      const partner: Partner = {
+        beneficiary: await wallet.getAddress(),
+        share: shares[index],
```
franciscotobar, line 37: There is a typescript error here regarding. `Type 'number | undefined' is not assignable to type 'number'.`

test/MultitokenCollector.test.ts
```diff
@@ -0,0 +1,409 @@
...
+  let fakeERC20Tokens: FakeContract<ERC20>[];
+  let partners: Partner[];
+  let remainderDestination: SignerWithAddress;
+  let partner1: SignerWithAddress;
+  let partner2: SignerWithAddress;
+  let partner3: SignerWithAddress;
+  let partner4: SignerWithAddress;
+
+  beforeEach(async function () {
+    [owner, partner1, partner2, partner3, partner4, remainderDestination] =
```
franciscotobar, line 89: `Type 'SignerWithAddress | undefined' is not assignable to type 'SignerWithAddress'.
  Type 'undefined' is not assignable to type 'SignerWithAddress'`

test/MultitokenCollector.test.ts
```diff
@@ -0,0 +1,409 @@
...
+        collector.updateRemainderAddress(Wallet.createRandom().address)
+      ).to.be.rejectedWith('There is balance to share');
+    });
+
+    it('Should reject non-owner calls', async function () {
+      const collector = await deployColector({});
+
+      const [newRemainderDestination, notTheOwner] = (
+        await ethers.getSigners()
+      ).slice<SignerWithAddress>(11, 13);
```
franciscotobar, line 400:
```suggestion
      ).slice(11, 13);
```
Is this valid?

test/MultitokenCollector.test.ts
```diff
@@ -0,0 +1,409 @@
...
+    it('Should update remainder address when any token balance is zero', async function () {
+      const collector = await deployColector({});
+      const expectedAddress = Wallet.createRandom().address;
+      await collector.updateRemainderAddress(expectedAddress);
+      const actualAddress = await collector.getRemainderAddress();
+
+      expect(actualAddress).eql(expectedAddress);
+    });
+
+    it(`Should withdraw all tokens' non-zero balances to the previous remainer address`, async function () {
```
franciscotobar, line 341:
```suggestion
    it(`Should withdraw all tokens' non-zero balances to the previous remainder address`, async function () {
```

test/MultitokenCollector.test.ts
```diff
@@ -0,0 +1,409 @@
...
+        collectorFactory.deploy(
+          owner.address,
+          fakeERC20Tokens.map(({ address }) => address),
+          partners,
+          remainderDestination.address
+        )
+      ).to.be.rejectedWith('Shares must add up to 100%');
+    });
+
+    it('Should not let deploy if the array of paMultitokenCollectorrtners is empty', async function () {
```
franciscotobar, line 181: Im not sure what are you trying to describe here.

jurajpiar: just accidentally pasted something there I believe. removing and also removing "let" for clatity

test/MultitokenCollector.test.ts
jurajpiar: Added
```
    /*
     * Adding `undefined` to array item type makes less sense 
     * in testing, where items of arrays in runtime are 
     * generally guaranteed to exist, thus we overwrite
     * `noUncheckedIndexedAccess` to false.
     */
    "noUncheckedIndexedAccess": false,
```
to tsconfig

test/MultitokenCollector.test.ts
jurajpiar: done

test/MultitokenCollector.test.ts
jurajpiar: done

test/MultitokenCollector.test.ts
jurajpiar: resolved with noUncheckedIndexedAccess

contracts/MultitokenCollector.sol
franciscotobar: Im not sure, I did some testing following the same link and its not necessary unless that we want to avoid sending extra gas, but over here is not possible. https://solidity-by-example.org/call/.  Therefore I think that we can remove it.

franciscotobar (approved):
LGTM

tsconfig.json
```diff
@@ -8,6 +8,13 @@
      * due to import of types from typechain-types.
     */
     "importsNotUsedAsValues": "remove",
+    /*
+     * Adding `undefined` to array item type makes less sense 
+     * in testing, where items of arrays in runtime are 
+     * generally guaranteed to exist, thus we overwrite
+     * `noUncheckedIndexedAccess` to false.
+     */
+    "noUncheckedIndexedAccess": false,
```
franciscotobar, line 17: maybe we need to set this back to true in the `tsconfig.build.json`

jurajpiar: @rsksmart/payments what do you think?

contracts/MultitokenCollector.sol
jurajpiar: ok will do then

jurajpiar: done! [ref](https://github.com/rsksmart/rif-relay-contracts/compare/72ebd56c074df9948ca62ba5f2b85b1e2b875fd2..b3a71a903eee79722f83118614f2af6cda6ff946)

tsconfig.json
franciscotobar: I think that its ok, but lets see if someone else has something to say.

antomor: Instead of disabling this globally, I'd have preferred to force the type where necessary

test/Collector.test.ts
```diff
@@ -0,0 +1,526 @@
...
+        partners,
+        remainderDestination.address
+      );
+
+      const actualOwnerAddress = await collector.owner();
+      const actualTokens = await collector.getTokens();
+
+      expect(actualOwnerAddress, 'Failed to set the owner').to.equal(
+        expectedOwnerAddress
+      );
```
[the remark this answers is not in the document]
antomor: It applies to all the expects

test/Collector.test.ts
```diff
@@ -0,0 +1,526 @@
...
+
+      await expect(
+        collectorFactory.deploy(
+          owner.address,
+          fakeERC20Tokens.map(({ address }) => address),
+          partners,
+          remainderDestination.address
+        )
+      ).to.be.rejectedWith('0 is not a valid share');
+    });
```
antomor, line 201: Shouldn't we add a test case to check the deployment if no tokens are specified?

test/Collector.test.ts
```diff
@@ -0,0 +1,526 @@
...
+  });
+
+  describe('updateShares', function () {
+    let newPartners: Partner[];
+    let deployColector: CollectorDeployFunction;
```
antomor, line 206:
```suggestion
    let deployCollector: CollectorDeployFunction;
```

test/Collector.test.ts
```diff
@@ -0,0 +1,526 @@
...
+      deployColector = prepareDefaultDeployer(
+        collectorFactory,
+        owner.address,
+        fakeERC20Tokens.map(({ address }) => address),
+        partners,
+        remainderDestination.address
+      );
+    });
+
+    it('Should withdraw for each parther', async function () {
```
antomor, line 416:
```suggestion
    it('Should withdraw for each partner', async function () {
```

test/Collector.test.ts
```diff
@@ -0,0 +1,526 @@
...
+        6
+      ) as SignerWithAddress;
+
+      await collector.transferOwnership(newOwnerAddress);
+
+      expect(await collector.owner()).to.be.equal(newOwnerAddress);
+    });
+
+    it('Should fail when is called by an address that is not the owner', async function () {
+      // const [, , , , , , newOwner, notTheOwner] = await hardhat.getSigners();
```
antomor, line 515: could we remove this line?

test/Collector.test.ts
```diff
@@ -0,0 +1,526 @@
...
+      const { address: newOwnerAddress } = (await ethers.getSigners()).at(
+        6
+      ) as SignerWithAddress;
+
+      await collector.transferOwnership(newOwnerAddress);
+
+      expect(await collector.owner()).to.be.equal(newOwnerAddress);
+    });
+
+    it('Should fail when is called by an address that is not the owner', async function () {
```
antomor, line 514:
```suggestion
    it('Should fail when called by an address that is not the owner', async function () {
```

antomor (changes requested):
@jurajpiar Thank you for the changes, they look good apart from some minor comments I left.
Furthermore, I think we should change [deploy-collector.input.sample.json](https://github.com/rsksmart/rif-relay-contracts//blob/b15dc108cc74061f5463a2f60b144c73435c7eb4/deploy-collector.input.sample.json) and also [the analysis document](https://www.notion.so/iovlabs/Multi-token-Collector-Proposal-b60250405f24489da4537491f22861e9) to reflect the new changes.

tsconfig.json
jurajpiar: Sure, however that requires changes in 14 files as this causes 
a

antomor:
>  while executing the npm run test, Im getting the following warning Failed to generate 2 stack traces. Run Hardhat with --verbose to learn more., before this warning was not showing.


@jurajpiar do we know the reason behind it?

antomor:
Furthermore, I think that we should add the functions to add/remove tokens from the collector?

contracts/Collector.sol
```diff
@@ -56,46 +58,47 @@ contract Collector is ICollector {
...
+        return _tokens;
     }
 
-    function withdraw() external override onlyOwner {
+    function getRemainderAddress() external view returns (address) {
+        return _remainderAddress;
+    }
+
+    function withdrawToken(IERC20 token) public onlyOwner {
         uint256 balance = token.balanceOf(address(this));
```
ulasanil, line 102: Would it be good idea to check if the `token` is in the `_tokens` array before withdrawing? Currently, `token` could be any ERC20 address. Since this function can only called by owner, I don't see much of an impact but still wanted to raise this issue.

test/Collector.test.ts
jurajpiar: As discussed, no.

test/Collector.test.ts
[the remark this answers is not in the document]
jurajpiar: as discussed this order is one that we defined.

contracts/Collector.sol
jurajpiar: I am not convinced that it would be worth the extra computation/gas.

contracts/Collector.sol
```diff
@@ -9,45 +9,47 @@ import "@openzeppelin/contracts/math/SafeMath.sol";
...
-        require(totalShares == 100, "Shares must add up to 100%");
-        _;
-    }
-
     modifier onlyOwner() {
         require(msg.sender == owner, "Only owner can call this");
         _;
     }
 
     modifier noBalanceToShare() {
```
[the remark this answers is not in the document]
jurajpiar: This only adds the percentages, not the actual values.

test/Collector.test.ts
jurajpiar: done

test/Collector.test.ts
jurajpiar: done

test/Collector.test.ts
jurajpiar: done

test/Collector.test.ts
jurajpiar: done

test/tasks/deployCollector.test.ts
```diff
@@ -111,6 +112,7 @@ describe('Deploy Script', function () {
       });
 
       it('deploy a collector', async function () {
+        //FIXME: Mismatch of test case and test itself
```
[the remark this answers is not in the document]
jurajpiar: did we talk about creating a task to go through the tests and fix where necessary @antomor ? I can't remember...

tsconfig.json
jurajpiar: As discussed, we leave it set to false for now, as it would require changes in 14 files. Should we create task for this @antomor ?

contracts/Collector.sol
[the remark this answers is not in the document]
ulasanil: Sorry Github noob here. I meant the `noBalanceToShare` modifier. If I am not mistaken, following `require` would fail if there is any token left to share. Attacker can send small amount of tokens to the Collector contract to make this condition fail deliberately. Of course, the risk is.... questionable. 

```solidity
    require(
                _tokens[i].balanceOf(address(this)) < _partners.length,
                "There is balance to share"
            );
```

tsconfig.json
antomor: Done, https://rsklabs.atlassian.net/browse/PP-678

contracts/Collector.sol
antomor: @ulasanil Although it may make sense, I don't see any risk related to this, is there?

ulasanil: It would just prevent unintended external calls to other smart contracts I guess. But I don't see a security risk. I think we can ignore this.

antomor (changes requested):

contracts/Collector.sol
```diff
@@ -90,10 +90,24 @@ contract Collector is ICollector {
...
 
+    function addToken(IERC20 token) external onlyOwner {
+        _tokens.push(token);
+    }
+
     function getTokens() external view returns (IERC20[] memory) {
         return _tokens;
     }
 
+    function removeToken(uint256 tokenIndex) external onlyOwner {
```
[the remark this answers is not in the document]
jurajpiar: hm... true.
The only solution I can think of right now would be:
`removeToken(uint256 tokenIndex, address tokenAddress)` ... `require(_tokens[tokenIndex] == tokenAddress, "Incorrect token index")`. wdyt @antomor ?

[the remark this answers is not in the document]
antomor: We decided to extend the [`TokenHandler`](https://github.com/rsksmart/rif-relay-contracts/blob/main/contracts/TokenHandler.sol) for consistency

contracts/Collector.sol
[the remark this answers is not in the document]
antomor: @ulasanil thank you for pointing it out. We do agree that there's a risk here, but due to its level, we'll mitigate it in a separate task.

ulasanil (approved):
LGTM!

contracts/Collector.sol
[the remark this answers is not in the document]
jurajpiar: @antomor I realised a couple of issues with this approach while implementing it.
1. removeToken would not check for existing balance (the main issue really)
2. uses `require(tokens[token] == true, "Token is not accepted");` which isn't necessary (nor wanted potentially) in Collector
3. related to the above - requires constant update of the tokens mapping (misused for single purpose)
4. we already discussed it, but for completeness - separate ownable implementation  - using it however requires changing the Ownable contract property `address private _owner` to `internal` to use it in the Collactor's constructor

All-and-all it appears to be less attractive as a solution, now. wdyt?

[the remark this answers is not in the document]
jurajpiar: I'd probably go back to the original impl and just add that check to the removeToken

ironFe93:
@antomor added a token address parameter for the removeToken() function, fixed tests and made changes to the input file and deploy file. Also modified the document analysis.

test/Collector.test.ts
```diff
@@ -0,0 +1,734 @@
...
+
+      expect(actualOwnerAddress, 'Failed to set the owner').to.equal(
+        expectedOwnerAddress
+      );
+      expect(actualTokens.toString(), 'Failed to set tokens').to.equal(
+        expectedTokens.toString()
+      );
+    });
+
+    it('Should deploy a Collector with multiple tokens', async function () {
```
franciscotobar, line 147: I think that we could include the unit testing if we are going to do a deployment without tokens.

antomor: Thank you for highlighting it, done

tasks/deployCollector.ts
```diff
@@ -79,11 +79,17 @@ export const deployCollector = async (
   },
   {});
 
+  const collectorTokens = await collector.getTokens();
+  const tokensToPrint = {};
+  collectorTokens.forEach((tokenAddress, i) => {
+    Object.assign(tokensToPrint, { [`Collector Token ${i}`]: tokenAddress });
+  });
```
jurajpiar, line 86: I'd do this immutably:
```suggestion
const tokensToPrint = (await collector.getTokens()).reduce( (acc, token, i) => {
  acc[`Collector Token ${i}`] = tokenAddress;
  
  return acc;
} , {});
```

test/Collector.test.ts
```diff
@@ -0,0 +1,757 @@
...
+
+      await collector.withdrawToken(token.address);
+
+      expect(token.transfer).to.have.callCount(expectedTransferCount);
+    });
+
+    it('Should fail when not enough revenue to share', async function () {
+      const tokens = fakeERC20Tokens.map((token) => {
+        return token.address;
+      });
```
jurajpiar, line 677:
```suggestion
      const tokens = fakeERC20Tokens.map(({address}) => address);
```

ironFe93: done

test/Collector.test.ts
```diff
@@ -0,0 +1,757 @@
...
+      expect(token.transfer).to.have.callCount(expectedTransferCount);
+    });
+
+    it('Should fail when not enough revenue to share in any of the tokens', async function () {
+      const tokens = fakeERC20Tokens.map((token, i) => {
+        const isLastToken = fakeERC20Tokens.length - 1;
+        token.balanceOf.returns(partners.length - (i === isLastToken ? 1 : 0));
+
+        return token.address;
+      });
```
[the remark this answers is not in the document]
ironFe93: done

tasks/deployCollector.ts
ironFe93: done

antomor (approved):
LGTM

contracts/Collector.sol
antomor: Implemented

contracts/Collector.sol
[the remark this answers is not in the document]
antomor: We couldn't extend the tokenHandler, so decide to implement something similar.

PR #121, opened 2023-02-14
Automate the contracts deployment

david-iov:
## What

- Apart from the server register step, this is the only step that is performed manually. We want to automate this. In the same way we did for the other libraries.

## Why

- Currently still a manual process.

### DIFF
--- a/.github/workflows/contracts-cd.qa.yml
+++ b/.github/workflows/contracts-cd.qa.yml
@@ -11,17 +11,17 @@ jobs:
     environment:
       name: QA
 
-      steps:
-        - uses: actions/checkout@v3
-        - name: Use Node.js
-          uses: actions/setup-node@v2
-          with:
-            node-version: "16.x"
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: "16.x"
 
-        - name: Install dependencies
-          run: npm install
+      - name: Install dependencies
+        run: npm install
 
-        - name: Deploy Contract on Testnet
-          run: |
-            MNEMONIC=${{ secrets.QA_MNEMONIC }}  npx hardhat deploy --network testnet
-            cat contract-addresses.json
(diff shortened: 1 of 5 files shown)


.github/workflows/contracts-cd.qa.yml
```diff
@@ -0,0 +1,32 @@
+---
+name: Deploy Contracts CD for QA
+
+on:
+  push:
+    branches: [main, PP-675/contracts-cd]
```
antomor, line 6: We need to remove the test branch
```suggestion
    branches: [main]
```

antomor (approved):
LGTM

.github/workflows/contracts-cd.rwm.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: "16.x"
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Deploy Contract on Testnet
+        run: |
+          MNEMONIC='${{ secrets.QA_MNEMONIC }}'  npx hardhat deploy --network testnet
```
antomor, line 27: I think the secret name should be different, shouldn't it?

.github/workflows/contracts-cd.rwm.yml
```diff
@@ -0,0 +1,33 @@
...
+      
+      - name: Upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: contract-addresses.json
```
antomor, line 32: it there the possibility that one environment file overwrite an existing one?

antomor (changes requested):
@david-iov Thank you for the changes, I think we should use different environment variable. In the meantime, I'm going to change the hardhat config to include mainnet in the networks.

.github/workflows/contracts-cd.rwm.yml
antomor: mainnet network configuration added

antomor (approved):
LGTM

PR #124, opened 2023-02-17
PP-678/noUncheckedIndexAccess

franciscotobar:
## What

- Enable the property noUncheckedIndexAccess in the tsconfig.json.

## Why

- Avoid having it globally disabled

## Refs

- (PP-678)[https://rsklabs.atlassian.net/browse/PP-678]


antomor (approved):
LGTM

PR #126, opened 2023-03-13
PP720: Modify withdraw script to allow it to run with just one token

AndresQuijano:
## What

- Modify withdraw script so it receives an additional optional parameter representing the token address to withdraw. If not sent, all allowed tokens will be withdrawn.

## Why

- Adapt the script to the recent Collector changes.

## Refs

- [PP-720](https://rsklabs.atlassian.net/browse/PP-720)


antomor (approved):
LGTM

PR #129, opened 2023-03-20
PP-670/smart-wallet-factory-migration

franciscotobar:
## What

- Migrate SmartWalletFactory unit testing

## Why

- Part of the web3 to ethers migration

## Refs

- [PP-670](https://rsklabs.atlassian.net/browse/PP-670)

### DIFF
--- a/test/factory/smartWalletFactory.test.ts
+++ b/test/factory/smartWalletFactory.test.ts
@@ -1,5 +1,10 @@
 import { expect } from 'chai';
-import { SmartWallet, SmartWalletFactory, UtilToken } from 'typechain-types';
+import {
+  SmartWallet,
+  SmartWalletFactory,
+  SmartWalletFactory__factory,
+  UtilToken,
+} from 'typechain-types';
 import { ethers } from 'hardhat';
 import { constants, utils, Wallet } from 'ethers';
 import { createValidPersonalSignSignature } from '../utils/createValidPersonalSignSignature';
@@ -10,6 +15,7 @@ import {
   TypedDeployRequestData,
 } from '../utils/EIP712Utils';
 import { getSuffixData, HARDHAT_CHAIN_ID } from '../smartwallet/utils';
+import { deployContract } from '../../utils/deployment/deployment.utils';
 
 const minIndex = 0;
 const maxIndex = 1000000000;
@@ -17,13 +23,9 @@ const maxIndex = 1000000000;
 const nextIndex = () =>
   Math.floor(Math.random() * (maxIndex - minIndex + 1) + minIndex);
 
-const deployContract = <T>(contract: string) => {
-  return ethers
-    .getContractFactory(contract)
-    .then((contractFactory) => contractFactory.deploy() as T);
-};
-
-const provider = ethers.provider;
+type SmartWalletFactoryOptions = Parameters<
+  SmartWalletFactory__factory['deploy']
+>;
 
 describe('SmartWalletFactory', function () {
   describe('constructor', function () {
@@ -32,8 +34,13 @@ describe('SmartWalletFactory', function () {
 
     beforeEach(async function () {
       template = ethers.Wallet.createRandom();
-      const factory = await ethers.getContractFactory('SmartWalletFactory');
-      smartWalletFactory = await factory.deploy(template.address);
+      ({ contract: smartWalletFactory } = await deployContract<
+        SmartWalletFactory,
+        SmartWalletFactoryOptions
+      >({
+        contractName: 'SmartWalletFactory',
+        constructorArgs: [template.address],
+      }));
     });
 
     it('should update master copy', async function () {
@@ -48,9 +55,17 @@ describe('SmartWalletFactory', function () {
     let owner: Wallet;
 
     beforeEach(async function () {
-      const factory = await ethers.getContractFactory('SmartWalletFactory');
-      const template: SmartWallet = await deployContract('SmartWallet');
-      smartWalletFactory = await factory.deploy(template.address);
+      const { contract: template } = await deployContract<SmartWallet, []>({
+        contractName: 'SmartWallet',
+        constructorArgs: [],
+      });
+      ({ contract: smartWalletFactory } = await deployContract<
+        SmartWalletFactory,
+        SmartWalletFactoryOptions
+      >({
+        contractName: 'SmartWalletFactory',
+        constructorArgs: [template.address],
+      }));
       owner = ethers.Wallet.createRandom();
     });
 
@@ -70,10 +85,6 @@ describe('SmartWalletFactory', function () {
             recoverer,
             index
           );
-
-        await expect(
-          provider.getCode(smartWalletAddress)
-        ).to.eventually.be.equal('0x');
 
         const dataToSign = utils.solidityKeccak256(
           ['address', 'address', 'address', 'uint256'],
@@ -123,7 +134,7 @@ describe('SmartWalletFactory', function () {
 
         await expect(
           smartWalletFactory.createUserSmartWallet(
-            owner.address,
+            constants.AddressZero,
             recoverer,
             index,
             signature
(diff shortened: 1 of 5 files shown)


test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+
+describe('SmartWalletFactory', function () {
+  describe('constructor', function () {
+    let smartWalletFactory: SmartWalletFactory;
+    let template: Wallet;
+
+    beforeEach(async function () {
+      template = ethers.Wallet.createRandom();
+      const factory = await ethers.getContractFactory('SmartWalletFactory');
+      smartWalletFactory = await factory.deploy(template.address);
```
antomor, line 36: Why don't we use the `deployContract` function here? if it's because of the arguments, we could use the `deployContract` function defined in `deployment.utils.ts`

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+  getLocalEip712DeploySignature,
+  TypedDeployRequestData,
+} from '../utils/EIP712Utils';
+import { getSuffixData, HARDHAT_CHAIN_ID } from '../smartwallet/utils';
+
+const minIndex = 0;
+const maxIndex = 1000000000;
+
+const nextIndex = () =>
+  Math.floor(Math.random() * (maxIndex - minIndex + 1) + minIndex);
```
antomor, line 18: I don't like the fact that the next index depends completely on a random number, hence the tests are not repeatable. If possible, I'd avoid it or replace it with a [random generator that accepts a seed](https://github.com/davidbau/seedrandom).

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+          index
+        );
+        [worker, otherCaller] = await ethers.getSigners();
+        token = await deployContract('UtilToken');
+      });
+
+      it('should initialize the smart wallet in the expected address without paying fee', async function () {
+        const initialWorkerBalance = await token.balanceOf(worker.address);
+
+        expect(initialWorkerBalance).to.be.equal(0);
```
antomor, line 186: Instead of performing this check, I'd rather check that the final balance is equal to the initial one. No need for them to be always zero, do they?

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+          );
+
+        const smartWallet = await ethers.getContractAt(
+          'SmartWallet',
+          smartWalletAddress
+        );
+
+        const finalWorkerBalance = await token.balanceOf(worker.address);
+
+        expect(finalWorkerBalance).to.be.equal(0);
```
antomor, line 239: Instead of performing this check, I'd rather check that the final balance is equal to the initial one. No need for them to be always zero, do they?

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+          );
+
+        const smartWallet = await ethers.getContractAt(
+          'SmartWallet',
+          smartWalletAddress
+        );
+
+        const finalWorkerBalance = await token.balanceOf(worker.address);
+
+        expect(finalWorkerBalance).to.be.equal(amountToBePay);
```
antomor, line 303: Instead of testing that the initial worker balance is 0, I'd check that the final balance is equal to the initial balance plus the amount paid

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+        const initialWorkerBalance = await token.balanceOf(worker.address);
+
+        expect(initialWorkerBalance).to.be.equal(0);
+        await expect(
+          provider.getCode(smartWalletAddress)
+        ).to.eventually.be.equal('0x');
+
+        await token.mint(1000, smartWalletAddress);
+
+        const amountToBePay = 500;
```
antomor, line 253:
```suggestion
        const amountToPay = 500;
```
or 
```suggestion
        const amountToBePaid = 500;
```

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+        const finalWorkerBalance = await token.balanceOf(worker.address);
+
+        expect(finalWorkerBalance).to.be.equal(amountToBePay);
+        await expect(smartWallet.isInitialized()).to.eventually.be.true;
+      });
+
+      it('should fail with tokenGas equals to zero paying fee', async function () {
+        const initialWorkerBalance = await token.balanceOf(worker.address);
+
+        expect(initialWorkerBalance).to.be.equal(constants.Zero);
```
antomor, line 310: I'd remove this check

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+        await expect(
+          provider.getCode(smartWalletAddress)
+        ).to.eventually.be.equal('0x');
+
+        await token.mint(1000, smartWalletAddress);
+
+        const amountToBePay = 500;
```
antomor, line 317:
```suggestion
        const amountToPay = 500;
```
or
```suggestion
        const amountToBePaid = 500;
```

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+
+      it('should initialize the smart wallet in the expected address paying fee', async function () {
+        const initialWorkerBalance = await token.balanceOf(worker.address);
+
+        expect(initialWorkerBalance).to.be.equal(0);
+        await expect(
+          provider.getCode(smartWalletAddress)
+        ).to.eventually.be.equal('0x');
+
+        await token.mint(1000, smartWalletAddress);
```
antomor, line 251: Can we move this in the `beforeEach` function?

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,557 @@
...
+        await expect(smartWallet.isInitialized()).to.eventually.be.true;
+      });
+
+      it('should initialize the smart wallet in the expected address paying fee', async function () {
+        const initialWorkerBalance = await token.balanceOf(worker.address);
+
+        expect(initialWorkerBalance).to.be.equal(0);
+        await expect(
+          provider.getCode(smartWalletAddress)
+        ).to.eventually.be.equal('0x');
```
antomor, line 249: Do we need this check? Doesn't the function fail if the SW is already deployed?

antomor (changes requested):
@franciscotobar Overall the changes look good, I'd remove some parts of the tests if they're not necessary.

test/factory/smartWalletFactory.test.ts
franciscotobar: I can use apply this, but then I think that needs to be applied to the rif-relay testing as-well.

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: This is the only change that is pending, I don't see whats the issue about depending on a random number, if we don't have any type of dependency on this, during the tests.

test/factory/smartWalletFactory.test.ts
```diff
@@ -1,497 +1,585 @@
...
+        const smartWalletAddress =
+          await smartWalletFactory.getSmartWalletAddress(
+            owner.address,
+            recoverer,
+            index
+          );
+
+        await expect(
+          provider.getCode(smartWalletAddress)
+        ).to.eventually.be.equal('0x');
```
antomor, line 93: I'd remove this check also

antomor (changes requested):

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

test/factory/smartWalletFactory.test.ts
franciscotobar: updated

antomor (approved):
LGTM

PR #134, opened 2023-07-18
fix(PP-804): do not show an error when the address file is missing

antomor:
## What

- do not raise an error when the address file is missing during the contracts deployment

## Why

- the error can be misleading and the developer could think that the deployment wasn't successful

### DIFF
--- a/tasks/deploy.ts
+++ b/tasks/deploy.ts
@@ -13,10 +13,10 @@ export const writeConfigToDisk = (config: NetworkConfig) => {
   console.log(`Address file available at: "${ADDRESS_FILE}"`);
 };
 
-export const updateConfig = (
+export const updateConfig = async (
   contractAddresses: ContractAddresses,
   { hardhatArguments, config: { networks } }: HardhatRuntimeEnvironment
-): NetworkConfig => {
+): Promise<NetworkConfig> => {
   console.log('Generating network config...');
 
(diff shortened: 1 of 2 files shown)


tasks/deploy.ts
```diff
@@ -32,8 +33,15 @@ export const updateConfig = (
     throw new Error('Unknown Chain Id');
   }
 
+  let existingConfig = {};
+  try {
+    existingConfig = parseJsonFile<AddressesConfig>(ADDRESS_FILE);
+  } catch (error) {
+    console.log(`Previous configuration not found at: "${ADDRESS_FILE}"`);
+  }
```
jurajpiar, line 41:
```suggestion
  const existingConfig = await Promise.resolve(parseJsonFile<AddressesConfig>(ADDRESS_FILE)).catch(e => {
  console.log(`Previous configuration not found at: "${ADDRESS_FILE}"`);
  })
```
and changing `updateConfig` signature to `async` and calling it in `‎test/tasks/deploy.test.ts‎` (the only found usage) with `await`.

jurajpiar (approved):
LG, just would promisify the try/catch

tasks/deploy.ts
antomor: Updated to replace the try/catch with a promise

franciscotobar:
I tried it and it works properly :)

PR #136, opened 2023-09-19
PP-840: get allowed tokens for given verifiers only

antomor:
## What

- change the allowed-tokens to receive the verifiers to check
- change the allow-tokens to receive the verifiers to modify

## Why

- to make the script more flexible and avoid asking for unnecessary data

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -90,7 +90,7 @@ npx hardhat collector:change-partners --collector-address "<collector_address>" 
 ```
 Example:
 ```bash
-npx hardhat collector:change-partners --collector-address "<0x9b91c655AaE10E6cd0a941Aa90A6e7aa97FB02F4" --partner-config "partner-shares.json" --gas-limit "200000" --network regtest
+npx hardhat collector:change-partners --collector-address "0x9b91c655AaE10E6cd0a941Aa90A6e7aa97FB02F4" --partner-config "partner-shares.json" --gas-limit "200000" --network regtest
 ```
 
 #### Add collector token
--- a/Readme.md
+++ b/Readme.md
@@ -161,7 +161,7 @@ Once the smart contracts are deployed, tokens must be individually allowed to be
     - `<NETWORK_NAME>` is an optional parameter for the network name, taken from the `hardhat.config.ts` file (default value is `hardhat`) **important! This should be the same network name as the one used to deploy the contracts** 
 If you want to modify just one or few verifiers, please specify it with the argument `--verifier-list` in a comma-separated format: `npx hardhat allow-tokens --network <NETWORK_NAME> --token-list <TOKEN_ADDRESSES> --verifier-list <address1,address2>`
 2. To query allowed tokens run `npx hardhat allowed-tokens --network <NETWORK_NAME>`. This will display them on the console. If you want to query just one or few verifiers, please specify it with the argument `--verifier-list` in a comma-separated format: `npx hardhat allowed-tokens --network <NETWORK_NAME> --verifier-list <address1,address2>`
-3. To remove a sepecific token, run `npx hardhat remove-tokens --network <NETWORK_NAME> --token-list <TOKEN_ADDRESSES>` where:
+3. To remove a specific token, run `npx hardhat remove-tokens --network <NETWORK_NAME> --token-list <TOKEN_ADDRESSES>` where:
     - `<TOKEN_ADDRESSES>` is a comma-separated list of the token addresses to be removed on the available verifiers.
     - `<NETWORK_NAME>` is an optional parameter for the network name, taken from the `hardhat.config.ts` file (default value is `hardhat`) **important! This should be the same network name as the one used to deploy the contracts**.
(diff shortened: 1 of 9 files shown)


test/tasks/allowTokens.test.ts
```diff
@@ -5,15 +5,15 @@ import fs from 'fs';
 import * as hre from 'hardhat';
 import { ethers } from 'hardhat';
 import sinon from 'sinon';
-import { allowTokens } from '../../tasks/allowTokens';
+import { AllowedTokensArgs, allowTokens } from '../../tasks/allowTokens';
 import { stubReadFileSync } from './utils';
 
 use(chaiAsPromised);
 
 describe('Allow Tokens Script', function () {
```
franciscotobar, line 13: We are missing some testing, ie:

- should allow a list of tokens and verifiers
- should allow a single token and verifiers
- should allow a list of tokens and a single verifier

test/tasks/removeTokens.test.ts
```diff
@@ -7,13 +7,14 @@ import { ethers } from 'hardhat';
 import sinon from 'sinon';
 import { removeTokens } from '../../tasks/removeTokens';
 import { stubReadFileSync } from './utils';
+import { AllowedTokensArgs } from 'tasks/allowTokens';
 
 use(chaiAsPromised);
 
 describe('Remove Tokens Script', function () {
```
franciscotobar, line 14: We are missing some testing, ie:

- should remove a list of tokens and verifiers
- should remove a single token and verifiers
- should remove a list of tokens and a single verifier

test/tasks/allowTokens.test.ts
antomor: I totally agree with you, and actually, I had also added those tests at some point, but then I had to revert the changes because `sinon-chai` (the library used to spy on some functions) isn't compatible with `smock`, hence I reverted the changes 😓

hardhat.config.ts
```diff
@@ -71,6 +74,7 @@ const config: HardhatUserConfig = {
       ...sharedNetworkConfig,
       url: 'https://public-node.testnet.rsk.co',
       chainId: 31,
+      gasPrice: 40000,
```
franciscotobar, line 77: I'm not sure about setting the gas price here, if it's necessary for testnet, it's not going to be the same for mainnet? 

And setting this gas price could not affect if the tx can or It's going to be included in a block?

antomor: I had to set the `gasPrice` because without it, I was receiving the following error: `transaction's gas price exceeds cap`. I think this is one of the features that were introduced in the recent versions of RSKj. Since this is related to the interaction with the RSKj node, I thought this should have been a global configuration.
I haven't changed that value in mainnet because I haven't had the need to interact with it. If we experience the same problem (very likely), we may add it later on. WDYT?

tasks/removeTokens.ts
```diff
@@ -1,50 +1,27 @@
-import { ContractTransaction } from 'ethers';
```
franciscotobar, line 1: What is the expectation of the removal?

I did some testing, and it fails in some scenarios, but I'm not sure if are covered. 

For ie:

- Remove tokens that are not allowed
- Remove tokens that are in a specific verifier without specifying it

Readme.md
```diff
@@ -156,13 +156,15 @@ This file also is being exported on the distributable version to provide the con
...
+2. To query allowed tokens run `npx hardhat allowed-tokens --network <NETWORK_NAME>`. This will display them on the console. If you want to query just one or few verifiers, please specify it with the argument `--verifier-list` in a comma-separated format: `npx hardhat allowed-tokens --network <NETWORK_NAME> --verifier-list <address1,address2>`
+3. To remove a sepecific token, run `npx hardhat remove-tokens --network <NETWORK_NAME> --token-list <TOKEN_ADDRESSES>` where:
     - `<TOKEN_ADDRESSES>` is a comma-separated list of the token addresses to be removed on the available verifiers.
     - `<NETWORK_NAME>` is an optional parameter for the network name, taken from the `hardhat.config.ts` file (default value is `hardhat`) **important! This should be the same network name as the one used to deploy the contracts**.
+If you want to modify just one or few verifiers, please specify it with the argument `--verifier-list` in a comma-separated format: `npx hardhat remove-tokens --network <NETWORK_NAME> --token-list <TOKEN_ADDRESSES> --verifier-list <address1,address2>`
```
franciscotobar, line 167: We should standardize <network_name>, <NETWORK_NAME>, <NETWORK>

hardhat.config.ts
franciscotobar: I agree, we may need to wait to understand the behavior

Readme.md
antomor: good point, just used `NETWORK_NAME` everywhere 👍🏼

tasks/removeTokens.ts
franciscotobar: resolved in a discussion with @antomor

PR #147, opened 2023-11-21
refactor(task/withdraw): wait for transaction receipt

franciscotobar:
## What

- Wait for the transaction receipt while withdrawing funds from the collector.

## Why

- The previous balance was the same as the balance after the withdrawal, since we were not waiting the tx to be processed.

### DIFF
--- a/tasks/withdraw.ts
+++ b/tasks/withdraw.ts
@@ -84,9 +84,11 @@ export const withdraw = async (
   );
 
   try {
-    tokenAddress
+    const transaction = tokenAddress
       ? await collector.withdrawToken(tokenAddress, { gasLimit })
       : await collector.withdraw({ gasLimit });
+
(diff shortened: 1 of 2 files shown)


antomor (approved):
@franciscotobar Thank you for the changes. LGTM!

tasks/withdraw.ts
```diff
@@ -84,9 +84,12 @@ export const withdraw = async (
   );
 
   try {
-    tokenAddress
+    const transaction = tokenAddress
       ? await collector.withdrawToken(tokenAddress, { gasLimit })
       : await collector.withdraw({ gasLimit });
+
+    console.log(`Waiting for transaction ${transaction.hash}...`);
+    await transaction.wait();
```
antomor, line 92: I've just added a console.log to show the transaction hash.

PR #151, opened 2024-02-13
PP-891: modularize the contracts deployment

antomor:
## What

- Change the deployment script to receive flags
- Add the documentation to deploy only some contracts

## Why

- This allows the operator to make modular deployment instead of deploying everything


tasks/deployers.ts
```diff
@@ -0,0 +1,159 @@
...
+};
+export const deployRelayHub = async (
+  ethers: HardhatEthersHelpers
+): Promise<Pick<ContractAddresses, 'RelayHub' | 'Penalizer'>> => {
+  const relayHubF = await ethers.getContractFactory('RelayHub');
+  const penalizerF = await ethers.getContractFactory('Penalizer');
+  const { address: penalizerAddress } = await penalizerF.deploy();
+  const { address: relayHubAddress } = await relayHubF.deploy(
+    penalizerAddress,
+    1,
```
franciscotobar, line 149: I never think about this before, but what about if we want to change the parameters for the relayHub, should we hardcode it?

tasks/deploy.ts
```diff
@@ -2,19 +2,36 @@ import { HardhatEthersHelpers, HardhatRuntimeEnvironment } from 'hardhat/types';
...
+  versionRegistry?: boolean;
+};
+
 // TODO: Use the async version of fs.writeFile
 export const writeConfigToDisk = (config: NetworkConfig) => {
   fs.writeFileSync(ADDRESS_FILE, JSON.stringify(config));
   console.log(`Address file available at: "${ADDRESS_FILE}"`);
 };
 
 export const updateConfig = async (
```
franciscotobar, line 33: I tried to update the config after deploying multiple contracts, and if you deploy less contracts it does not erase the previous that are extra.

tasks/deployers.ts
antomor: Yes, I've realised this while doing this change. It's always been hard-coded and I don't like it either. I think in future we could provide an additional script just for deploying the relayHub, that will receive all the parameters required instead of hard-coding it 🤦🏽‍♂️

franciscotobar (approved):
LGTM

PR #154, opened 2024-03-21
fix(RREL-024): remove invalid comments

antomor:
## What

- remove invalid comments

## Why

- they can be misleading and bring unclarity


franciscotobar (approved):
LGTM

PR #155, opened 2024-04-11
refactor(BoltzVerifier): interaction for validation

franciscotobar:
## What

Fix the following issues:
- RREL-020
- RREL-021
- RREL-022

## Why

- To improve the quality of the code

## Refs

- [PP-905](https://rsklabs.atlassian.net/browse/PP-905)


contracts/utils/BoltzValidator.sol
```diff
@@ -0,0 +1,62 @@
...
+        } else if (method == _PUBLIC_METHOD) {
+            claim = abi.decode(
+                relayRequest.request.data[4:],
+                (NativeSwap.PublicClaimInfo)
+            );
+        } else {
+            revert("Method not allowed");
+        }
+
+        require(
```
antomor, line 43: Do we still need this? If so, I'd move it before the claim call

test/verifier/boltzDeployVerifier.test.ts
antomor, line 1: apart from the called method (external vs public), it looks like the tests are identical so I'd try to avoid repetition

antomor (approved):
@franciscotobar The code looks good to me

contracts/verifier/BoltzDeployVerifier.sol
```diff
@@ -43,55 +43,41 @@ contract BoltzDeployVerifier is
...
+            amount = claim.amount;
+        }
 
-        if (relayRequest.request.tokenAmount > 0) {
-            if (relayRequest.request.tokenContract != address(0)) {
-                require(
-                    tokens[relayRequest.request.tokenContract],
-                    "Token contract not allowed"
-                );
+        if (relayRequest.request.tokenContract != address(0)) {
```
antomor, line 62: If tokenAmount is 0, we don't need this check at all, do we?

contracts/verifier/BoltzRelayVerifier.sol
```diff
@@ -61,26 +61,22 @@ contract BoltzRelayVerifier is
             "Destination contract not allowed"
         );
 
-        if (relayRequest.request.tokenAmount > 0) {
-            if (relayRequest.request.tokenContract != address(0)) {
-                require(
-                    tokens[relayRequest.request.tokenContract],
-                    "Token contract not allowed"
-                );
+        if (relayRequest.request.tokenContract != address(0)) {
```
antomor, line 64: If `tokenAmount` is 0, we don't need this check at all, do we?

contracts/utils/BoltzValidator.sol
antomor: Apparently, we don't really need it

contracts/verifier/BoltzRelayVerifier.sol
antomor: we don't need to check for the token amount

antomor (approved):
LGTM

ilanolkies (approved):

PR #165, opened 2024-08-02
Boltz verifier claim amount

franciscotobar:
## What

- Boltz verifier validates claiming amount in native payment

## Why

- To avoid executing a relay tx that cannot cover the payment


antomor (approved):
LGTM

raullaprida (approved):
Minor comments. Looks Good overall

contracts/verifier/BoltzDeployVerifier.sol
```diff
@@ -48,25 +49,33 @@ contract BoltzDeployVerifier is
...
+            } else {
+                uint256 amount = BoltzUtils.validateClaim(
+                    relayRequest.request.data,
+                    relayRequest.request.to,
+                    contractAddr
+                );
+
+                require(
+                    relayRequest.request.tokenAmount <=
+                        address(contractAddr).balance + amount,
```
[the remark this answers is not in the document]
antomor: yes, we didn't want to refactor all the variables because this would have been a breaking change, so where possible we renamed it, while here it's still `tokenAmount` even if it refers to RBTC

contracts/verifier/BoltzRelayVerifier.sol
```diff
@@ -56,27 +57,35 @@ contract BoltzRelayVerifier is
...
-                relayRequest.request.tokenAmount <= address(payer).balance,
-                "Native balance too low"
-            );
+                require(
+                    relayRequest.request.tokenAmount <=
+                        address(payer).balance + amount,
+                    "Native balance too low"
+                );
+            }
         }
```
VaWheel, line 89: what if `relayRequest.request.tokenAmount` is 0? should we allow it or revert?

franciscotobar: If `tokenAmount` it will not go in into this flow since it is validated in line 62.

PR #168, opened 2024-09-01
estimation-no-signature

franciscotobar:
## What

- Contract execution validation in boltz smart wallet initialization

## Why

- During estimation process without signature it is required to simplify the process

### DIFF
--- a/contracts/smartwallet/BoltzSmartWallet.sol
+++ b/contracts/smartwallet/BoltzSmartWallet.sol
@@ -94,6 +94,7 @@ contract BoltzSmartWallet is BaseSmartWallet {
 
         bool success;
         bytes memory ret;
+        // Although this check isn't strictly necessary, it's included to improve the transaction estimation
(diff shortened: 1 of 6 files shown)


contracts/smartwallet/BoltzSmartWallet.sol
```diff
@@ -92,11 +92,15 @@ contract BoltzSmartWallet is BaseSmartWallet {
...
         _setOwner(owner);
 
-        (bool success, bytes memory ret) = to.call{value: value}(data);
-        if (!success) {
-            if (ret.length == 0) revert("Unable to execute");
-            assembly {
-                revert(add(ret, 32), mload(ret))
+        bool success;
+        bytes memory ret;
+        if (to != address(0)) {
```
antomor, line 97: I'd add a comment here to clarify that it's not strictly necessary, but it's helpful at estimation time.

antomor (approved):
LGTM

PR #170, opened 2024-09-02
estimation-no-signature

franciscotobar:
## What

- Contract execution validation in minimal boltz smart wallet initialization

## Why

- During estimation process without signature, it is required to simplify the process


antomor (approved):
LGTM

